PR #1, opened 2020-05-27
Add Tree -> DataFrame functionality

Moelf:
This is mainly a wrapper / short cut thing, but I noticed the following:

It would be nice to have syntax such as `f["t1] |> DataFrame`, or something like, `f["t1"]["float_array"] |> collect` (acting like an iterator) but currently the `IOStream` info is in `f` so we always need to pass `f` as an argument. I'm not sure what's a better design for this since in python it was natural to `f.tree.branch.array()` and you would have the file information in parent object.


tamasgal:
Nice, thanks for the PR!

Since we are in a very early stage of the API design, we can of course change that and propagate the stream pointer to subtrees and branches, that's of course doable.

My only concern is how to deal with nested (jagged) arrays, since those will not work with `DataFrames`. Not sure how `uproot` is doing that since I do not use that feature (most of our data is jagged), but some meaningful error messages should be enough I guess ;)

In future I'd like to use `Requires.jl` to lazy load functions when other libraries (like `DataFrames.jl`) are available, without depending on them directly to keep `UnROOT.jl` slim.

Project.toml
```diff
@@ -5,18 +5,19 @@ version = "0.1.3"
 
 [deps]
 CodecZlib = "944b1d66-785c-5afd-91f1-9de20f533193"
+DataFrames = "a93c6f00-e57d-5684-b7b6-d8193f3e46c0"
 MD5 = "6ac74813-4b46-53a4-afec-0b5dc9d7885c"
 Mixers = "2a8e4939-dab8-5edc-8f64-72a8776f13de"
 Parameters = "d96e819e-fc66-5662-9728-84c9c7592b0a"
 StaticArrays = "90137ffa-7385-5640-81b9-e52037218182"
 
 [compat]
```
tamasgal, line 14: Is the resorting intentional or does it come from some IDE? 😉

Moelf: I did the `] activate .` then `add` thing

tamasgal: The comment was targeting the reordering in `compat` (MD5.jl etc.), but never mind, I was just curious ;)

Moelf: yeah, I think the re-ordering was also done by `Pkg`, I'm on `1.4.1`

Moelf:
> how to deal with nested (jagged) arrays

naturally, we can't make non-flat .root file into dataframe, I should implement a check probably, should be easy (like flat-ness check). Regarding jagged array, I think there should be a separate package to provide a wrapper interface to `Tables.jl`, so this `Awkward.jl` can deal with general jagged array, and interface with `UnROOT`.

This `Awkward.jl` should be much lighter than python counterpart since
1) we don't worry about using for-loop nearly as much
2) there's no pure vs. numpy `data type` issue
The `AwkwardArray` should just be a wrapper around `AbstractArray` and providing certain friendly syntax, since weather or not the underlying array is `lazy` or not or even on `GPU` we don't need to worry.

Project.toml
tamasgal: Alright, then only the `using DataFrames` in `runtests.jl` is missing to pass the CI!

tamasgal:
Yes I am constantly thinking about how to implement the jaggedness. At this very moment the first prototypes are handled by `UnROOT.splitup()` which will take the raw data (`Vector{UInt8}`) and offset information (coming from `UnROOT.array(f, path; raw=true)`) and split the data into a vector of vectors.

To be honest, Julia can happily deal with `Vector{Vector{}}` so there are cases where there is no need for any "awkwardness", but of course having that fancy indexing and all the other nice features you mentioned (GPU) is a thing, so there will definitely be an interface to `Awkward.jl` or similar packages.

PR #17, opened 2021-06-30
fix one hack, more to go; add reference

Moelf:


tamasgal:
Thanks for getting this rolled again ;)

Issue #4, opened 2020-05-29
Redesign version management of streamers


tamasgal:
The current design of the streamer type hierarchy (including versioning) works in a way which I do not find nice at all. At some point I thought it would be a nice solution but now I see a lot of code repetitions.

A lot of streamer logic is currently implemented in `bootstrap.jl` and the overall implementation of a streamer and all of its versions is done like this:

```julia
abstract type TAttLine <: ROOTStreamedObject end
struct TAttLine_1 <: TAttLine end
function readfields!(io, fields, T::Type{TAttLine_1})
    fields[:fLineColor] = readtype(io, Int16)
    fields[:fLineStyle] = readtype(io, Int16)
    fields[:fLineWidth] = readtype(io, Int16)
end
struct TAttLine_2 <: TAttLine end
function readfields!(io, fields, T::Type{TAttLine_2})
    fields[:fLineColor] = readtype(io, Int16)
    fields[:fLineStyle] = readtype(io, Int16)
    fields[:fLineWidth] = readtype(io, Int16)
end
```

As seen above, the `ROOTStreamedObject` is the supertype (which is just a simple abstract type) and there is (always) one single subtype (another abstract type) representing a specific streamer.

Different versions are defined by appending `_VERSION` to the type name and creating an empty struct. The `readfields!` method then define how to read its fields.

Both versions of `TAttLine` (v1 and v2) have the same fields with the same types, so in this case it's just code repetition.

Note that this definitions should be created dynamically in future, but the overall layout is of course the same.

During the read-in of a ROOT file, UnROOT checks if there is a streamer with the given name and version (appended), like `TAttLine_2` and if not, it will complain. (later it will define it from the streamer info).

A simplification of this system which allows a nice bootstrapping (no code repetition) and a better version management would be something like the following:

```julia
import Base: getproperty

abstract type ROOTStreamedObject end

function Base.getproperty(obj::ROOTStreamedObject, sym::Symbol)
    if !haskey(Core.getproperty(obj, :fields), sym)
        error("Type $(typeof(obj)) has no field $(String(sym))")
    end
    Core.getproperty(obj, :fields)[sym]
end

struct TAttLine{V} <: ROOTStreamedObject  # this should be a macro
    fields::Dict{Symbol, Any}
end
```

where the actual struct definition could of course be simplified via a macro like `@rootstreamer TAttLine` or so.

The `readfields!()` methods can then be defined using value type dispatch:

```julia
[...]

PR #18, opened 2021-06-30
[RFC] streamer type generation

Moelf:
I think your original `TODO` had a typo? we are generating for `readfields!` it seems.

We come up with a structure so we don't need to type out an entire `@eval function ....` in the `_generate_T_Types`. This PR serves as a POC, and at least avoid hard-coding for each n in `TType_n` variation

### DIFF
--- a/src/bootstrap.jl
+++ b/src/bootstrap.jl
@@ -1,13 +1,20 @@
 # A collection of bootstrapped code which should be generated
 # dynamically in future.
 
-abstract type TNamed <: ROOTStreamedObject end
-struct TNamed_1 <: TNamed end
-function readfields!(io, fields, ::Type{TNamed_1})
-    parsefields!(io, fields, TObject)
-    fields[:fName] = readtype(io, String)
-    fields[:fTitle] = readtype(io, String)
+# TODO: furthuer factoring this out, maybe parsing some ROOT headers
+function _generate_T_Types(name)
+    if startswith(string(name), "TName")
+        @eval function readfields!(io, fields, ::Type{$name})
+            parsefields!(io, fields, TObject)
+            fields[:fName] = readtype(io, String)
+            fields[:fTitle] = readtype(io, String)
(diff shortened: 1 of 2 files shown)


tamasgal:
Yes, there were a few (changing) naming conventions throughout.

Some of the `Type_n` things will be need though for bootstrapping, i.e. to provide a fixed set of streamer definitions for well known types which are not described inside ROOT files.

I really need to sit down and think about how to restructure and come up with an easy to follow/understand naming convention for `read()`, `readfields()` and whatnot `;)`

tamasgal:
Sorry for spamming; for the sake of completeness I'd like to link our discussion from last year https://github.com/tamasgal/UnROOT.jl/issues/4

Moelf:
my question is there's no way to know about the ROOT built-in types such as `TName_1` and `_2` without looking at their source code right? at least for quite many basic types

tamasgal:
Yes, exactly. That's what we call "bootstrapping". I got most of these hard coded definitions from `uproot`, that's easier to read compared to the C++ code 😁

Moelf:
will we ever encounter TTypes whose name we have never seen before? (i.e. not in ROOT base) We can't generate `readfields!` in that case because we know nothing about the struct right? (it's self documented by using ROOT base TTypes somehow?)

tamasgal:
Yes, I think those types are always provided "in source" and not "in file".

The stuff in `uproot` is spread out. I created a modified version of `uproot3` back then, and dumped the classes which were generated during the runtime. Jim made it so that it basically builds a string from all the streamer information from building blocks and the string is then passed to `eval()`, basically containing the full class definition.
My modified `uproot3` version always created a `*.py` with the name of the class and the version of it. So I could do a `uproot.open("whatever.root")` and had a folder with all the e.g. `TNamed_3.py`, `TTree_2.py`, `KM3NET_...py`, etc. ;)

PR #19, opened 2021-07-01
Compression algo: LZ4 and LZMA at least for Streamers and tests

Moelf:


Moelf:
Eventually we need to make an iterative approach for this anyways because data apparently cannot be dumped easily..... (idk why, but I thought LZMA should be as tolerant as LibZ but no, it doesn't like extra garbage bytes at the end even if we're not even reading into them)

tamasgal:
Amazing, thanks!

PR #22, opened 2021-07-02
Support Jagged branch from CMS NanoAOD

Moelf:
- handle multiple blocks in datastream
- hard code TLeafB


Moelf:
```
julia> using UnROOT

julia>  array(ROOTFile("tree_with_jagged_array.root"), "t1/int32_array"; raw=true)
(Int32[0, 0, 1, 0, 1,


In [5]: up.open("tree_with_jagged_array.root")["t1"]["int32_array"].array()
Out[5]: <JaggedArray [[] [0] [0 1] ... 
```

the offsets is a bit off (i.e. needs manual add subtract), will fix tmr

Moelf:
@tamasgal I'm gonna merge this since tests and converge both seem happy. I'd propose drop 1.0 support since next LST (1.6 or 1.7) is right on the corner.

tamasgal:
Very nice work!

I'd suggest to add `hasproperty` since this seems to be the only requirement we need and it's a one-liner: https://github.com/JuliaLang/julia/blob/6aaedecc447e3d8226d5027fb13d0c3cbfbfea2a/base/reflection.jl#L1583-L1590

Something like below right into `UnROOT.jl` or in a separate file `compat.jl`:

```julia
if VERSION < v"1.2"
    hasproperty(x, s::Symbol) = s in fieldnames(typeof(x))
end
```

PR #24, opened 2021-07-05
[WIP] Huntdown allocation and type instability

Moelf:
- clean up streamer decompression
- reduce Preamble type instability
- fVersion typo
- improve TBasketKey type stability


src/root.jl
```diff
@@ -271,52 +296,33 @@ end
...
 
-    s = datastream(io, basketkey)  # position(s) == 0, but offsets start at -basketkey.fKeylen
-    start = position(s)
-    contentsize = basketkey.fLast - basketkey.fKeylen
     offsetbytesize = basketkey.fObjlen - contentsize - 8
-    offset_len = offsetbytesize ÷ 4 # these are always Int32
+    offsetnumints = offsetbytesize ÷ 4 # these are always Int32
+    l1 = contentsize + 4 # see the graph above
 
+    data = @view basketrawbytes[begin:contentsize]
```
giordano, line 311: `begin` isn't in Julia v1.0.  You may have to use `firstindex(basketrawbytes)` instead

Moelf: I changed to hard code `1`. I doubt we'd support 1.0 in our 1.0 release, the shared file system is un-usable without the tarball registry, and also pre-1.5 TTFP is exacerbated by poor hardware.

Moelf:
for posterity, a common benchmark I run is:
```
julia> const rf = ROOTFile("./test/samples/NanoAODv5_sample.root")
julia> @benchmark UnROOT.arrays(rf, "Events")
BechmarkTools.Trial: 25 samples with 1 evaluations.
 Range (min … max):  179.443 ms … 278.009 ms  ┊ GC (min … max):  0.00% … 12.65%
 Time  (median):     199.686 ms               ┊ GC (median):    10.95%
 Time  (mean ± σ):   203.733 ms ±  18.769 ms  ┊ GC (mean ± σ):  10.57% ±  2.71%

           ▅▅  █                                                 
  ▅▁▁▁▁▁▁█▁██▅██▅▁▁▁▁▁▅▁▅▅▁▁▁▁▁▁▁▁▁▁▅▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▅ ▁
  179 ms           Histogram: frequency by time          278 ms <

 Memory estimate: 147.42 MiB, allocs estimate: 1892035.
```

which will hopefully shed light on why we're allocating so much in the case where all branches are cached.

PR #28, opened 2021-07-07
Add Iterator interface fix #6

### LINKED ISSUE #6: Iterator and Table interface
This allows smaller memory footprint by avoiding materialize the entire array into memory. And we can use something like [OnlineStats](https://github.com/joshday/OnlineStats.jl)

 I tried to implement one for a few days now and find `fBasketSeek` is very annoying and want some suggestion. I'm not sure what would be the canonical way to do this but:
```
struct BRANCH_ITR
    io
    seeks
    thetype
    count::Int
end

function Base.iterate(S::BRANCH_ITR, idx=1)
    basket_seek = S.seeks[idx]
    if idx > S.count
        return nothing
    elseif basket_seek==0
        return (nothing, idx+1)
    else
        s = datastream(io, basketkey)
        return (readtype(s, S.thetype), idx)
    end
end

```

this doesn't work because `readtype` advance the cursor and we lost track of it in the next iteration, where would you suggest to keep track of:
- basket cursor
- cursor within the basket?

https://quinnj.home.blog/2020/11/13/partition-all-the-datas/ Leverage this as ROOT provides natural partition too.

Moelf:
The code is a bit messy but the idea is there. I will clean up a bit before merging just want to know if people have thoughts on the implementation. The speed seems okay, allocation probably can be reduced, but a revamp on the `interped_data()` would benefit all use case.

image

It's thread-safe to iterator through multiple branches by simply zipping them together:

```
julia> rf = ROOTFile("./tree_with_large_array.root")
ROOTFile("./tree_with_large_array.root") with 1 entry and 17 streamers.

julia> it1 = BranchItr(rf, rf["t1/int32_array"])
itBranchItr{Int32, UnROOT.Nojagg}
Branch: int32_array
Total entry: 100000

julia> it2 = BranchItr(rf, rf["t1/float_array"])
BranchItr{Float32, UnROOT.Nojagg}
Branch: float_array
Total entry: 100000

julia> for (a,b) in zip(it1,it2)
           a, b
       end
```


Moelf:
@aminnj mentioned that in ROOT when you iterate over and entire tree, one should be lazy about basket unpacking because maybe there's a `continue` branch that is met 99.99% of the time (so you can skip all other 100 branches for almost all the time).

We don't have a tree iterator interface yet and this current design is hard to fit the bill, any suggestion would be welcomed.

tamasgal (approved):

Moelf:
image
I combined the indexing and iteration interface, such that we support broadcast and map over a `LazyBranch`

tamasgal:
Looks very good!

PR #29, opened 2021-07-08
Add tests for custom KM3NeT stuff

tamasgal:
This stuff will probably removed in a later stage but currently it's good to have in the code base for crosschecks.

### DIFF
--- a/src/io.jl
+++ b/src/io.jl
@@ -11,6 +11,7 @@ Base.position(c::Cursor) = position(c.io)
 
 
 function unpack() end
+packedsizeof(T::Type) = sum(sizeof.(fieldtypes(T)))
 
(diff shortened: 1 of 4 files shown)


src/io.jl
```diff
@@ -11,6 +11,7 @@ Base.position(c::Cursor) = position(c.io)
 
 
 function unpack() end
+packedsizeof(T::Type) = sum(sizeof.(fieldtypes(T)))
```
Moelf, line 14: does `Base.summarysize()` give correct size here?

tamasgal: No, that includes some padding, unfortunately.

tamasgal:
@all-contributors please add @tamasgal for code, doc, infra, data and tests

PR #30, opened 2021-07-08
Tree iteration and table

Moelf:
ehhh this seems to just work, can some one test this with a large enough root file that would definitely crash session if somewhere in it is not lazy?

```
rf = ROOTFile()
tb = Table(rf, treename);
```


Moelf:
it seems all lazy already :|

There are two way to iterate:
One is:
```
for row in table
    row.electron_pt, blah...
end
```
where `row` is materialized each iteration,

alternatively, one can do:
```
for i in eachindex(table)
    table.MET[i] < 150 && continue
end
```
this way if user puts easy to fail cut up front, we simply don't materialize all the other baskets in other branches.

I believe this is all we need because I looked at `Table <: AbstractArray` and iteration is simply doing indexing, this is so nice.

Moelf:
The problem with always using `TypedTable` was that the compiler just get crushed. The current implementation seems okay even with a massive 1700 branches tree:
```

julia> const t = @time ROOTFile("./test/samples/NanoAODv5_sample.root");
  1.851695 seconds (2.18 M allocations: 129.126 MiB, 3.43% gc time, 99.77% compilation time)

julia> table = @time Table(t, "Events")
  4.765739 seconds (11.25 M allocations: 1.166 GiB, 2.54% gc time, 91.66% compilation time)
UnROOT.LazyTree:
  NumColumns: 1479
  NumEntry: 1000

julia> @time table[1].Electron_dxy
  0.951131 seconds (1.87 M allocations: 112.183 MiB, 2.66% gc time, 99.73% compilation time)
1-element Vector{Float32}:
 0.00037050247

julia> @time table.Electron_dxy[1]
  0.017181 seconds (48 allocations: 3.203 KiB, 99.82% compilation time)
1-element Vector{Float32}:
 0.00037050247
```

Issue #23, opened 2021-07-03
Reading subbranches of jagged custom struct


8me:
I try to read a `TBranchElement` which contains subbranches from a data file which has the file format like [this example file](https://git.km3net.de/km3py/km3net-testdata/-/blob/master/km3net_testdata/data/offline/km3net_offline.root). In order to get more specific, I want to read the track information at `E/Evt/trks`. The track information contains the particles properties (for simulation) or the reconstructions (for measurement) per event, so this leads to a nested array structure. Taking a look at it using `uproot`  (loading the given example file to `fobj`) I get:
```
In [14]: fobj["E/Evt/trks"].show()
name                 | typename                 | interpretation                
---------------------+--------------------------+-------------------------------
trks                 | vector<Trk>              | AsGroup(<TBranchElement 'trks'
trks.fUniqueID       | uint32_t[]               | AsJagged(AsDtype('>u4'))
trks.fBits           | uint32_t[]               | AsJagged(AsDtype('>u4'))
trks.usr_data        | std::vector<AAny>*       | AsObjects(AsArray(True, Fal...
trks.usr_names       | std::vector<std::stri... | AsObjects(AsArray(True, Fal...
trks.id              | int32_t[]                | AsJagged(AsDtype('>i4'))
trks.pos.x           | double[]                 | AsJagged(AsDtype('>f8'))
trks.pos.y           | double[]                 | AsJagged(AsDtype('>f8'))
trks.pos.z           | double[]                 | AsJagged(AsDtype('>f8'))
trks.dir.x           | double[]                 | AsJagged(AsDtype('>f8'))
trks.dir.y           | double[]                 | AsJagged(AsDtype('>f8'))
trks.dir.z           | double[]                 | AsJagged(AsDtype('>f8'))
trks.t               | double[]                 | AsJagged(AsDtype('>f8'))
trks.E               | double[]                 | AsJagged(AsDtype('>f8'))
trks.len             | double[]                 | AsJagged(AsDtype('>f8'))
trks.lik             | double[]                 | AsJagged(AsDtype('>f8'))
trks.type            | int32_t[]                | AsJagged(AsDtype('>i4'))
trks.rec_type        | int32_t[]                | AsJagged(AsDtype('>i4'))
trks.rec_stages      | std::vector<int32_t>*    | AsObjects(AsArray(True, Fal...
trks.status          | int32_t[]                | AsJagged(AsDtype('>i4'))
[...]

PR #31, opened 2021-07-09
Reading subbranches of jagged custom struct

8me:
Attempt to solve the Issue #23 in order to read fields/subbranches of custom classes by using `fType` from the leafes.
"[WIP]": The determination of the jagged-offset is not 100% clear to me, because the reading of the jagged array from the given example in the mentioned issue requires no offset.

### DIFF
--- a/src/root.jl
+++ b/src/root.jl
@@ -169,21 +169,39 @@ function interped_data(rawdata, rawoffsets, branch, ::Type{J}, ::Type{T}) where 
 
 end
 
-# function interp_jaggT(branch, leaf)
+function _normalize_ftype(fType)
+    # Taken from uproot4; thanks Jim ;)
+    if Const.kOffsetL < fType < Const.kOffsetP
+        fType - kOffsetP
+    else
+        fType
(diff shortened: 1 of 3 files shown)


Moelf:
btw can we have a small test file for this?

8me:
@Moelf in the samples there is already a [test file](https://github.com/tamasgal/UnROOT.jl/blob/master/test/samples/km3net_offline.root). I also added a short test for one of the subbranches in the file, i.e. `E/Evt/trks/trks.t`.

tamasgal:
Nice!

tamasgal:
@8me can you rebase? afterwards we can merge if it's ready!

8me:
I'm right about to resolve the conflicts, I'm just not sure if the determination of the offsets is correct the way I reorganized it :grimacing:

Moelf:
I don't know the bytes magic involved but looks you've surgically added minimal code needed for the test to work and we have test now so LGTM! Thanks a lot!

PR #34, opened 2021-07-10
Add cache after branch is interpreted

aminnj:
There's currently a 3GB cache for raw unpacked basket data. Adding a 1GB cache after branch interpretation can help a lot for non-trivial branches especially when playing/exploring interactively on a small set of branches. E.g.,

```julia
julia> length(t)
962975

julia> t.Electron_pt[1:5]
5-element Vector{Vector{Float32}}:
 []
 [18.828014]
 [31.292875, 25.108507]
 [106.96358, 19.502377, 16.525389]
 [21.751484, 13.741731]
```
before (without 1GB cache)
```julia
julia> @time length.(t.Electron_pt);
  1.000231 seconds (4.82 M allocations: 324.212 MiB, 3.58% gc time)
```
after (with  1GB cache)
```julia
julia> @time length.(t.Electron_pt);
  0.138824 seconds (3.48 k allocations: 8.036 MiB)
```

### DIFF
--- a/src/iteration.jl
+++ b/src/iteration.jl
@@ -41,7 +41,7 @@ Reads an array from ith basket of a branch. Set `raw=true` to return raw data an
 """
 basketarray(f::ROOTFile, path::AbstractString, ithbasket) = basketarray(f, f[path], ithbasket)
 
-function basketarray(f::ROOTFile, branch, ithbasket)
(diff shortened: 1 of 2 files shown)


Moelf (approved):

tamasgal:
Thanks, that looks good to me!

We should think about how to expose the cache settings for the user, since this stuff is heavily experiment/use-case specific.

tamasgal (approved):

PR #35, opened 2021-07-11
Better pretty table

Moelf:
May be upstream-able https://github.com/JuliaData/TypedTables.jl/issues/81


tamasgal:
Nice! Looking forward to see this upstreamed. 

Should we add a simple loop over the sample files and their branches to verify that the printing always works?

Moelf:
I reworked this by making a real `LazyTree` struct this time and is `<: AbstractDataFrame` and internally still a `TypedTables.Table`. This has 2 merits:
1. we have more freedom and less worry when customizing behavior of LazyTree
2. we have access to a lot of tools in the DataFrame ecosystem

as an example of `1.`, now the `tree[1:5]` which materialize a typed `Table` is also displayed correctly
image


and my test shows that there's no visible performance difference:
```
julia> const r = ROOTFile("data18.root");

julia> const tree = LazyTree(r, "ntupOutput", ["jet_pt", "jet_eta"]);

julia> function mainloop(tree)
           ptsums = sizehint!(Float64[], length(tree))
           for row in tree
               best_η = 0.0
               id1 = id2 = -1
               for (i1, eta1) in enumerate(row.jet_eta)
                   for (i2, eta2) in enumerate(row.jet_eta)
                       i2 <= i1 && continue
                       Δη = abs(eta1-eta2)
                       if Δη > best_η
                           best_η = Δη
                           id1, id2 = i1, i2
                       end
                   end
               end
               ptsum = -1.
               if id1 != -1
                   ptsum = row.jet_pt[id1] + row.jet_pt[id2]
               end
               push!(ptsums, ptsum)
           end
           ptsums
       end
mainloop (generic function with 1 method)

julia> @time mainloop(tree);
  5.669738 seconds (9.05 M allocations: 1000.002 MiB, 5.05% gc time, 20.44% compilation time)

julia> @time mainloop(tree);
  2.623368 seconds (50.85 k allocations: 14.249 MiB)

julia> size(tree)
(635673, 2)
```

vs.

current `master`:
```

julia> @time mainloop(tree);
  5.896258 seconds (9.05 M allocations: 998.937 MiB, 4.79% gc time, 20.20% compilation time)

julia> @time mainloop(tree);
  2.775108 seconds (50.78 k allocations: 14.208 MiB)
```

PR #37, opened 2021-07-12
Thread-safety & locking

8me:
I propose to introduce the [exemplary](https://github.com/JuliaLang/julia/blob/master/base/lock.jl#L20-L25) usage of the `lock` at least in `_getindex()`. The reading of the baskets itself in  `readbasketseek()` is very low level and straightforward, so I guess the risk of something going wrong there is not too high.


Moelf:
LGTM, sorry for using a hacky pattern in the beginning.

PR #38, opened 2021-07-13
TLorentzVector support with improved user custom interface

### LINKED ISSUE #33: TLorentzVector branch support
@8me since you're looking at relevant code maybe I can borrow your wisdom! I'm trying to figure out the padding for TLorentz vector: https://github.com/scikit-hep/uproot3-methods/blob/ba9a97b3dc71c7030a9ec15a9d97397b5ff8aa0d/uproot3_methods/classes/TLorentzVector.py
```
import ROOT as r
from array import array

f = r.TFile("LorentzVector.root", "recreate")
t = r.TTree("t1","t1")

#x y z e
p4 = r.TLorentzVector(1,1,0,5)
t.Branch("LV.","TLorentzVector",p4)

for i in range(10):
     p4.SetPxPyPzE(1,1,0,i)
     t.Fill()
t.Write()
f.Close()
```

Moelf:
fix #33:
```julia
julia> r = ROOTFile("./test/samples/lorentzvector.root")
ROOTFile with 1 entry and 19 streamers.
./test/samples/lorentzvector.root
└─ t1
   ├─ "LV"
   └─ ""


julia> tree = LazyTree(r, "t1");

julia> tree[1].LV
LorentzVectors.LorentzVector{Float64}(4.0, 1.0, 2.0, 3.0)

julia> tree.LV[1]
LorentzVectors.LorentzVector{Float64}(4.0, 1.0, 2.0, 3.0)
```

(our normal tests are failing at the moment, looking into them)


Moelf:
The steps for hooking into the branch interpreting are:
1. user needs to provide a struct with the same name as `classname` at top level, in this example:
https://github.com/tamasgal/UnROOT.jl/pull/38/files#diff-287db97d7aaaee37763548eb2d18e567ca0ecd419ff119dafacc8bb194b0312cR33
2. provide a method given a branch and their type, what the resultant `eltype` of the branch should be and what is the "jaggness":
https://github.com/tamasgal/UnROOT.jl/pull/38/files#diff-287db97d7aaaee37763548eb2d18e567ca0ecd419ff119dafacc8bb194b0312cR42
3. given this `eltype` and jaggness, what to do with the raw bytes and offsets:
https://github.com/tamasgal/UnROOT.jl/pull/38/files#diff-287db97d7aaaee37763548eb2d18e567ca0ecd419ff119dafacc8bb194b0312cR45

tamasgal:
I am not sure if I am happy with implicit information delivery, regarding

> provide a struct with the same name as classname at top level

some class names are really nasty (see KM3NeT 😆) and others are totally generic, like `Hit` (see KM3NeT 🙈 ) or so 😉 

But overall, I agree that this approach is close to what I visualise for the 1.0 release!

Moelf:
ok I've polished the custom interface and all original + added tests are passing now.

@tamasgal I agree with what you said, I was gonna remove `eval` and change to get module but they are both implicit. Alternatively, we can allow user to provide a `Dict{String, Type}` mapping for use inside `interp_jaggT`. 

I like this way better

tamasgal:
Yes, I think a central (name)space would be nice to gather all the dynamically generates and custom definitions. Just like with a plug-in system where you register those?

Moelf:
I think custom struct collection per ROOTFile is probably enough. I will make `LazyBranch` and `LazyTree` have this keyword argument as well and default to `ROOTFiles`'s map. So that user don't have to re-make `ROOTFile` (just save some typing I guess)

src/custom.jl
```diff
@@ -16,12 +16,45 @@ function splitup(data::Vector{UInt8}, offsets, T::Type; skipbytes=0)
...
 
-# Custom, hardcoded streamers
-
+# Custom struct interpretation
 abstract type CustomROOTStruct end
 
-# KM3NeT
+reinterpret(vt::Type{Vector{T}}, data::AbstractVector{UInt8}) where T <: CustomROOTStruct = reinterpret(T, data)
+
+function interped_data(rawdata, rawoffsets, ::Type{J}, ::Type{T}) where {J <: JaggType, T <: CustomROOTStruct}
```
8me, line 24: Why is this forwarding needed for `T <: CustomROOTStruct`, I guess if there is no dispatch on the type like for `TLorenzVector` the basic `interped_data` from `root.jl` is called anyway, isn't it?

src/root.jl
```diff
@@ -47,9 +47,6 @@ function ROOTFile(filename::AbstractString)
     # Reading the header key for the top ROOT directory
     seek(fobj, header.fBEGIN + header.fNbytesName)
     dir_header = unpack(fobj, ROOTDirectoryHeader)
-    if dir_header.fSeekKeys == 0
-        ROOTFile(format_version, header, fobj, tkey, [])
```
Moelf, line 51: this just doesn't work anymore right? we don't have this constructor

Moelf:
continue in #41

Issue #39, opened 2021-07-13
Bug in offsets readout, unexpected fall back to 0


tamasgal:
I just discovered a serious bug in the `data` and `offset` readout while I was working with larger ROOT files. I think it's mostly affecting large datasets, but I am not sure.

With small sample file, we have no issues, everything is read out correctly and I can assemble the hits.

Just as a side note: a `UnROOT.KM3NETDAQHit` consists of 10 bytes and per event we have a 10 bytes header.

```julia
julia> f = ROOTFile("/home/tgal/Dev/UnROOT.jl/test/samples/km3net_online.root")ROOTFile with 10 entries and 54 streamers.
/home/tgal/Dev/UnROOT.jl/test/samples/km3net_online.root
├─ E
│  ├─ "Evt"
│  └─ ""
├─ KM3NET_TIMESLICE
│  ├─ "KM3NET_TIMESLICE"
│  └─ ""
├─ KM3NET_TIMESLICE_L0
│  ├─ "km3net_timeslice_L0"
│  └─ ""
├─ KM3NET_TIMESLICE_L1
│  ├─ "km3net_timeslice_L1"
│  └─ ""
├─ KM3NET_TIMESLICE_L2
│  ├─ "km3net_timeslice_L2"
│  └─ ""
├─ KM3NET_TIMESLICE_SN
│  ├─ "km3net_timeslice_SN"
│  └─ ""
├─ KM3NET_EVENT
│  ├─ "KM3NET_EVENT"
│  └─ ""
└─ KM3NET_SUMMARYSLICE
   ├─ "KM3NET_SUMMARYSLICE"
   └─ ""


julia> data, offsets = UnROOT.array(f, "KM3NET_EVENT/KM3NET_EVENT/snapshotHits"; raw=true)
(UInt8[0x40, 0x00, 0x03, 0xc6, 0x00, 0x09, 0x00, 0x00, 0x00, 0x60  …  0x30, 0x40, 0xa9, 0x7d, 0x0a, 0x3c, 0x21, 0xc9, 0x03, 0x17], Int32[0, 970, 2220, 3010])

julia> length(data)
3010

julia> length(offsets)
4

julia> offsets
4-element Vector{Int32}:
    0
  970
 2220
 3010

julia> sum(diff(offsets))
3010

julia> minimum(diff(offsets))
790

julia> UnROOT.splitup(data, offsets, UnROOT.KM3NETDAQHit, skipbytes=10)
[...]

PR #40, opened 2021-07-13
Preliminary fix of offset-readout

tamasgal:
This is a preliminary fix for #39. Comments are welcome (my time is very limited though :see_no_evil: )

### DIFF
--- a/src/root.jl
+++ b/src/root.jl
@@ -229,7 +229,8 @@ function readbranchraw(f::ROOTFile, branch)
         seek==0 && return
         data, offset = readbasketseek(f, branch, seek)
         append!(datas, data)
-        append!(offsets, (@view offset[begin+1:end]) .+ position)
+        # FIXME: assuming offset has always 0 or at least 2 elements ;)
+        append!(offsets, (@view offset[2:end]) .+ position)
         if length(offset) > 0
             position = offset[end]
         end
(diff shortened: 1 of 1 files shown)


src/root.jl
```diff
@@ -223,12 +223,14 @@ end
...
     datas = sizehint!(Vector{UInt8}(), sum(nbytes)) # maximum length if all data are UInt8
-    offsets = sizehint!(Vector{Int32}(), branch.fEntries+1) # this is always Int32
+    offsets = sizehint!(zeros(Int32, 1), branch.fEntries+1) # this is always Int32
+    position = 0
     foreach(branch.fBasketSeek) do seek
         seek==0 && return
         data, offset = readbasketseek(f, branch, seek)
         append!(datas, data)
-        append!(offsets, offset)
+        append!(offsets, offset[2:end] .+ position)
```
Moelf, line 232:
```suggestion
        append!(offsets, @view(offset, 2:end) .+ position)
```

Moelf:
LGTM

src/root.jl
tamasgal: Yep, thanks!

tamasgal:
The `zeros` are there for extra confusion, this  needs a bit of cleanup soon.

One test is failing, I guess assuming that `offset` will have at least 2 elements is too naive `;)`

tamasgal:
I have to figure out how to create a small file with many baskets though, we need a sample file for this.

Moelf:
how about `begin+1:end`

tamasgal:
I am just struggling with the `@view`  and the range containing `begin/end`

Moelf (approved):

PR #57, opened 2021-07-20
Add a logo

8me:


8me:
WIP because svg illustration differs for different browsers.

Moelf:
with the caveat that it doesn't look as good under dark mode but I blame GitHub

Issue #61, opened 2021-07-23
`leaf.fLenType==0` falsely classifying jagged branch as non-jagged


aminnj:
So I came across a super weird issue. I couldn't open up an analysis ntuple with a pretty basic type (`vector<float>`). Apparently this depends on the environment 😕. Test file:
```python
import ROOT as r

f = r.TFile("output_good.root", "recreate")
t = r.TTree("Events", "")

obj = r.vector("float")()
t.Branch("Jet_pt", obj)

rows = [[], [27.324586868286133, 24.88954734802246, 20.853023529052734], [], [20.330659866333008], [], []]
for row in rows:
    obj.clear()
    for x in row:
        obj.push_back(x)
    t.Fill()
t.Write()
f.Close()
```
gives me
```julia
julia> ROOTFile("output_good.root")["Events/Jet_pt"]
6-element LazyBranch{Vector{Float32}, UnROOT.Offsetjagg}:
 []
 [27.324587, 24.889547, 20.853024]
 []
 [20.33066]
 []
 []
```
if I make the ntuple on my laptop (MacOS, Python3, ROOT 6.16/00). But it fails with
```julia
julia> ROOTFile("output_bad.root")["Events/Jet_pt"]
6-element LazyBranch{Vector{Float32}, UnROOT.Nojagg}:
Error showing value of type LazyBranch{Vector{Float32}, UnROOT.Nojagg}:
ERROR: MethodError: Cannot `convert` an object of type Float32 to an object of type Vector{Float32}
```
if I make the ntuple on a remote machine in a CMS environment (SLC7/`CMSSW_10_2_5`, Python2.7 😬, ROOT 6.12/07).

Raw bytes are the same according to `array(...; raw=true)`. The only difference I found was that `fLeaves` on that branch has
```julia
  fLenType: Int32 4 # "bad"/remote ntuple
  fLenType: Int32 0 # "good"/local ntuple
```
which means that 
```julia
function JaggType(leaf)
    # https://github.com/scikit-hep/uproot3/blob/54f5151fb7c686c3a161fbe44b9f299e482f346b/uproot3/interp/auto.py#L144
    (match(r"\[.*\]", leaf.fTitle) !== nothing) && return Nooffsetjagg
    leaf isa TLeafElement && leaf.fLenType==0 && return Offsetjagg
    return Nojagg
end
```
[...]

PR #62, opened 2021-07-24
Pass streamers and branch into `JaggType` to fix issue 61

aminnj:
Closes https://github.com/tamasgal/UnROOT.jl/issues/61 . May not be elegant, but at least now *everything* (root file/streamers, branch, and leaf) is available inside the `JaggType` function for maximum flexibility to determine interpretation in the future.


tamasgal:
LGTM. We can make it "nicer" later, for now we focus on functionality anyways :) Thanks!

PR #63, opened 2021-07-24
check type stability in CI

aminnj:
Check that some example computations are fully type stable to avoid performance regression


tamasgal:
Thanks! Maybe we should add InteractiveUtils only for the tests as dependency

aminnj:
Done, I think

tamasgal (approved):

PR #65, opened 2021-07-25
use ArraysOfArrays to reduce allocation

### LINKED ISSUE #64: Poor performance when weaving many branches together
Sometimes in HEP the memory access pattern is inherently bad. For example we have jagged branches for electron and muon related quantities: Lorentz Vectors, isolation flags etc.

Let's say the `e_mask` is quality cut, and people want to extract isolation cuts for electrons that passed the `e_mask`:
```julia
function get_Isos(e_mask, m_mask, evt)
    v_l_passIso = Vector{Bool}[]

    v_e_passIso_HighPtCaloOnly = evt.v_e_passIso_HighPtCaloOnly
    v_e_passIso_TightTrackOnly_VarRad = evt.v_e_passIso_TightTrackOnly_VarRad
    v_e_passIso_TightTrackOnly_FixedRad = evt.v_e_passIso_TightTrackOnly_FixedRad
    v_e_passIso_Tight_VarRad = evt.v_e_passIso_Tight_VarRad
    #.... 12 more

    @inbounds for (i,f) in enumerate(e_mask)
        f || continue
        push!(
            v_l_passIso,
            Bool[
                v_e_passIso_HighPtCaloOnly[i],
                v_e_passIso_TightTrackOnly_VarRad[i],
                v_e_passIso_TightTrackOnly_FixedRad[i],
                v_e_passIso_Tight_VarRad[i],
                v_e_passIso_Loose_VarRad[i],
            ],
        )
    end
	# one more for loop for muons

	return v_l_passIso
```

which results in
image

Each of those piles is actually just the `v_e_passIso_HighPtCaloOnly = evt.v_e_passIso_HighPtCaloOnly` line.

This is probably horrible no matter what language you use, but I'm wondering if we should explore https://github.com/JuliaArrays/ArraysOfArrays.jl/ to see if we can cut down allocations of making concrete vectors (or vector of vectors) since in many cases they just get teared apart manually in the looper.

Moelf:
close #64


Moelf:
manually requesting @aminnj 's review with his many battle tested workload

aminnj:
Hmm, I tried this out on a 2.4M event Z->mumu tree. Uploaded [here](https://uaf-7.t2.ucsd.edu/~namin/dump/doublemu.root) (~400MB). I get 5x slower looping rate compared to `master`.
```julia
julia> const f = ROOTFile("doublemu.root") # 2.4M events

julia> const t = LazyTree(f, "t", [r"^Muon_(pt|eta|phi|mass)$","MET_pt"]);

julia> t.Muon_pt
2476431-element LazyBranch{Vector{Float32}, UnROOT.Nooffsetjagg}:
 [22.402422, 18.186892]
 [45.062744, 44.058678]
 [8.216898]
 ⋮
 [24.227955, 14.885574]
 [16.145634, 12.987432, 6.6345634, 4.7526903]

julia> struct DummyLV{T <: AbstractFloat}
    pt::T
    eta::T
    phi::T
    mass::T
end
```
```julia
julia> @time for (i,evt) in enumerate(t)
    length(evt.Muon_pt) < 2 && continue # at least 2 mu
    ((evt.Muon_pt[1] < 20) || (evt.Muon_pt[2] < 20)) && continue # leading 2 mu pt>20
    lvs = DummyLV.(evt.Muon_pt,evt.Muon_eta,evt.Muon_phi,evt.Muon_mass)
end

0.199550 seconds (511.96 k allocations: 56.539 MiB) # in master

1.161238 seconds (19.17 M allocations: 721.681 MiB, 11.49% gc time, 5.18% compilation time)
 # in this branch
```

Moelf:
hmm, I can imagine if loop is very tight already, getting views itself might be slower than having concrete numbers with very good locality.

aminnj:
Some more data. With this simpler test:
```julia
julia> @time for (i,evt) in enumerate(t)
    length(evt.Muon_pt) < 2 && continue # at least 2 mu
end
```
I got
### master
cold: 2.859309 seconds (24.56 M allocations: 1.072 GiB, 20.29% gc time, 0.16% compilation time)
warm: 0.030027 seconds (5.90 k allocations: 767.688 KiB)

### this PR
cold: 0.575812 seconds (4.92 M allocations: 373.350 MiB, 14.94% gc time, 1.65% compilation time)
warm: 0.234112 seconds (4.89 M allocations: 150.975 MiB, 7.24% gc time)

aminnj:
Time and allocations are reduced for cold runs in this PR wrt master, but subsequent runs are slow. I tried reverting the 3GB->1GB cache reduction and still see the slowness. So it must not be the cache.

Moelf:
I think I found the issue, 
before:
image

after:
image

those gaps ....

Moelf:
I wonder if we can pre-fetch the first basket of each `LazyBranch` at creation time which will help use infer the buffer type without too much user interference

Moelf:
do we want to merge this? or people have ideas for polishing. I think there's still some small instability issues, but since even with that `@assert`, none of the tests failed so I don't know what to look for....

Moelf:
@tamasgal do you know why CI is not working on this?

tamasgal (approved):

tamasgal:
Yes I think we should merge and then see how it performs in the wild 😜

PR #69, opened 2021-07-30
Reduce allocation by ~20%

Moelf:
With this I believe we can remove the LRU cache for interpred data because our LazyBranch cache + ArraysOfArrays is pretty good.

Basically if LRU cache for interpred data works super well, then one probably should just make concrete `tree` instead of having the same content in the form of LRU caches. I kept the raw data cache because there's nothing to re-use in-place there.

### DIFF
--- a/src/custom.jl
+++ b/src/custom.jl
@@ -38,7 +38,7 @@ Base.show(io::IO, lv::LorentzVector) = print(io, "LV(x=$(lv.x), y=$(lv.y), z=$(l
 function Base.reinterpret(::Type{LVF64}, v::AbstractVector{UInt8}) where T
     # first 32 bytes are TObject header we don't care
     # x,y,z,t in ROOT
-    v4 = ntoh.(reinterpret(Float64, v[1+32:end]))
+    v4 = ntoh.(reinterpret(Float64, @view v[1+32:end]))
     # t,x,y,z in LorentzVectors.jl
     LVF64(v4[4], v4[1], v4[2], v4[3])
 end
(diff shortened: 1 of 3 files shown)


Moelf:
before:
```julia
akako@ThinkX ~/Downloads> julia --startup-file no -e 'include("./demo.jl")'
  4.411549 seconds (13.79 M allocations: 2.653 GiB, 8.70% gc time, 22.15% compilation time)
```

after:
```julia
akako@ThinkX ~/Downloads> julia --startup-file no -e 'include("./demo.jl")'
  3.891949 seconds (15.00 M allocations: 2.321 GiB, 9.37% gc time, 21.29% compilation time)
```

on bigger work load,
before:
```
16.754587 seconds (45.82 M allocations: 6.171 GiB, 5.90% gc time, 2.82% compilation time)
```
after:
```
14.169935 seconds (46.21 M allocations: 5.706 GiB, 6.37% gc time, 2.96% compilation time)
```

tamasgal:
Ok nice one! But these enormous amounts of allocations are really awful. What does the demo.jo do?

Moelf:
for perspective, the second task in C++ allocates about 2.2G, but is significantly slower than Julia, likely because memory locality is poor in exchange for smaller memory foot print, I think it's not worth it. (C++ is like 3x slower or something, both cold run).

I think within 2x C++ allocation is fine, it's not like our peak memory usage is bad, this is cumulative allocation.

tamasgal:
Ah ok, that’s indeed good news then. Maybe we could try to get more coverage of the precompilation caching, but that’s probably a but tricky. I hope I find a few hours this weekend to work again on UnROOT, especially using this nice JET.ja package which was nicely advertised recently 😅

tamasgal (approved):

Moelf:
yes, I wouldn't make a release yet, not sure if this PR is robust enough tbh....

Issue #73, opened 2021-08-27
Working with LazyTree as a DataFrame


peremato:
I'm new to Julia wanting to see its potential for HEP computing. For this, I am playing with a TTree with UnROOT to see how this compares with equivalent job with ROOT/RDataFrame. 
I understand that a`LazyTree` is not a `DataFrame`, since in general you cannot not fit in memory the full `TTree`, however it provides the basic functionality of an `AbstractDataFrame`. Naively I thought that I could use initially a `LazyTree` and then once it is heavily reduced to be converted to a `DataFrame`. But this seems not to be working. Perhaps you can give me some hints.
The code I am trying to execute is as follows:
```julia
using UnROOT
f = ROOTFile("/eos/opendata/cms/derived-data/AOD2NanoAODOutreachTool/Run2012BC_DoubleMuParked_Muons.root")
events = LazyTree(f, "Events");
df = filter([:nMuon,:Muon_charge] => (n,c) -> n == 2 && c[1] != c[2], events[1:1000000])
```
and the error I get is:
```
[...]

PR #76, opened 2021-09-03
make branch basket buffer threadlocal

aminnj:
In each of the branch objects, the last-read basket data is stored. If one tried to naively `@threads` an event loop, there would be huge contention in this buffer cache. This PR makes `nthreads()` such buffer caches.

```julia
using UnROOT
using LVCyl
using FHist
using Base.Threads

const f = ROOTFile("Run2012BC_DoubleMuParked_Muons.root")
const t = LazyTree(f, "Events", ["nMuon", r"^Muon_(pt|eta|phi|mass|charge)"])

const hmass = Hist1D(Int64; bins=0:0.001:150)
@time @threads for (i,evt) in enumerate(t)
    evt.nMuon != 2 && continue
    sum(evt.Muon_charge) != 0 && continue
    lvs = LorentzVectorCyl.(evt.Muon_pt, evt.Muon_eta, evt.Muon_phi, evt.Muon_mass)
    mass = sum(lvs).mass
    push!(hmass, mass) # rename to atomic_push eventually
end
```
gives
```julia
27.897073 seconds (49.06 M allocations: 39.579 GiB, 23.23% gc time) # with @threads
68.897655 seconds (48.65 M allocations: 38.559 GiB, 7.48% gc time) # without @threads
```
for julia with 4 threads on a mac laptop. That's about a 2.5x speedup (consistent with me seeing 300% cpu usage instead of 400%).

Note that I commented out the LRU cache around `readbasketseek()` in both cases to keep RAM usage low (seems to go beyond the advertised `maxsize=1024^3` 🤷)

The example comes from https://github.com/tamasgal/UnROOT.jl/issues/73 .

### DIFF
--- a/src/UnROOT.jl
+++ b/src/UnROOT.jl
@@ -9,8 +9,10 @@ import AbstractTrees: children, printnode, print_tree
 
 using CodecZlib, CodecLz4, CodecXz, CodecZstd, StaticArrays, LorentzVectors, ArraysOfArrays
 using Mixers, Parameters, Memoization, LRUCache
+using Base.Threads
 
 import Tables, TypedTables, PrettyTables, DataFrames
+
 
 @static if VERSION < v"1.4"
     Base.first(a::AbstractVector{S}, n::Integer) where S<: AbstractString = a[1:(length(a) > n ? n : end)]
--- a/src/UnROOT.jl
+++ b/src/UnROOT.jl
@@ -9,10 +9,8 @@ import AbstractTrees: children, printnode, print_tree
 
 using CodecZlib, CodecLz4, CodecXz, CodecZstd, StaticArrays, LorentzVectors, ArraysOfArrays
 using Mixers, Parameters, Memoization, LRUCache
-using Base.Threads
 
 import Tables, TypedTables, PrettyTables, DataFrames
-
 
 @static if VERSION < v"1.4"
     Base.first(a::AbstractVector{S}, n::Integer) where S<: AbstractString = a[1:(length(a) > n ? n : end)]
--- a/src/iteration.jl
+++ b/src/iteration.jl
@@ -194,9 +194,9 @@ Base.firstindex(lt::LazyTree) = 1
 Base.lastindex(lt::LazyTree) = length(lt)
 
 # allow enumerate() to be chunkable (eg with Threads.@threads)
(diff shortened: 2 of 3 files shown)


tamasgal:
Awesome! Btw. the memory estimation is very basic and does not catch nested data well, that's why it can grow larger than the set 1GB.

aminnj:
Should we factor out the LRU cache for now so that people can turn it off without modifying the source?

```diff
 readbasket(f::ROOTFile, branch, ith) = readbasketseek(f, branch, branch.fBasketSeek[ith])

-@memoize LRU(; maxsize=1024^3, by=x -> sum(sizeof, x)) function readbasketseek(
+_cache_readbasketseek = LRU(; maxsize=1024^3, by=x -> sum(sizeof, x))
+@memoize _cache_readbasketseek function readbasketseek(
 # function readbasketseek(
 f::ROOTFile, branch::Union{TBranch, TBranchElement}, seek_pos::Int
 )::Tuple{Vector{UInt8},Vector{Int32}}  # just being extra careful
```
Based on https://github.com/JuliaCollections/Memoize.jl, I tried `memoize_cache(readbasketseek)` and couldn't get the object out. Maybe you know why.

aminnj:
Actually, I'll play around more with the cache and move my findings to the dedicated issue thread...

aminnj:
Since `@threads` relies on `getindex()` to do the chunking, this won't work:
```julia
@threads for (i,evt) in enumerate(t)
    i>5 && break
    evt.nMuon != 2 && continue
    sum(evt.Muon_charge) != 0 && continue
    # ...
end
```
since `enumerate(t)` doesn't assume a `getindex()` interface by default (see [here](https://github.com/JuliaLang/julia/pull/40788)), which is pretty inconvenient/unintuitive (until you realize how the macro works...). One would have to write
```julia
@threads for i in 1:length(t)
    t.nMuon[i] != 2 && continue
    sum(t.Muon_charge[i]) != 0 && continue
    # ...
end
```
However, our `LazyTable` in indexable, so based on the code [here](https://github.com/JuliaLang/julia/pull/40772), I added some extra definitions to allow us to `@threads ... enumerate(t)`.

aminnj:
As far as features, I think I'm done in this PR. A review is welcome :)

tamasgal (approved):

tamasgal:
Weeeha! 😃

Issue #78, opened 2021-09-04
Double reading of baskets near boundaries


aminnj:
For the file linked [here](https://github.com/tamasgal/UnROOT.jl/issues/73#issuecomment-907269846), I see weird behavior. When we iterate sequentially over one branch, we should be reading each basket only once. But what I see with an additional printout
```diff
diff --git a/src/iteration.jl b/src/iteration.jl
index b3579d5..2ac6b93 100644
--- a/src/iteration.jl
+++ b/src/iteration.jl
@@ -143,6 +143,7 @@ function Base.getindex(ba::LazyBranch{T,J,B}, idx::Integer) where {T,J,B}
     br = ba.buffer_range
     if idx ∉ br
         seek_idx = findfirst(x -> x > (idx - 1), ba.fEntry) - 1 #support 1.0 syntax
+        @show idx, br, seek_idx
         bb = basketarray(ba.f, ba.b, seek_idx)
         if typeof(bb) !== B
             error("Expected type of interpreted data: $(B), got: $(typeof(bb))")
```
is
```julia
julia> using UnROOT

julia> const t = LazyTree(ROOTFile("Run2012BC_DoubleMuParked_Muons.root"), "Events", ["Muon_pt"]);

julia> for i in 1:6000; t.Muon_pt[i]; end
(idx, br, seek_idx) = (1, 5477:7314, 1)
(idx, br, seek_idx) = (1810, 1:1809, 1)
(idx, br, seek_idx) = (1811, 1:1809, 2)
(idx, br, seek_idx) = (3647, 1811:3646, 2)
(idx, br, seek_idx) = (3648, 1811:3646, 3)
(idx, br, seek_idx) = (5476, 3648:5475, 3)
(idx, br, seek_idx) = (5477, 3648:5475, 4)
```
Each branch object has its own cache of the last-read basket and the associated index range `br`. If we `getindex()` with an index, we check if it's in `br` otherwise we have to read the whole basket.
`seek_idx` is the basket number. So we end up calling `basketarray()` (which does IO + decompression + interpretation) twice for each basket.

Note
```julia
julia> t.Muon_pt.b
...
fBasketEntry: Array{Int64}((524,)) [0, 1810, 3647, 5476, 7315, 9126, 10964, 12804, 14634, 16465  …  54231870, 55053565, 55875260, 56696955, 57518650, 58340345, 59162040, 59983735, 60805430, 61540413]
...
```

[...]

PR #79, opened 2021-09-04
prevent double basket reading

aminnj:
Closes https://github.com/tamasgal/UnROOT.jl/issues/78

We were unnecessarily reading baskets a second time when on the boundary of the next basket because `buffer_range` advertises one less element than we actually cache, so `getindex()` tries to re-read that entry.
```julia
julia> using UnROOT

julia> const t = LazyTree(ROOTFile("/Users/namin/sandbox/miscnotebooks/julia/Run2012BC_DoubleMuParked_Muons.root"), "Events", ["Muon_pt"])

julia> length(t.Muon_pt.buffer) # number of events we cache
734983

julia> length(t.Muon_pt.buffer_range) # number we advertise we cache
734982
```

### Before
```julia
julia> @time sum(length.(t.Muon_pt))
  4.812074 seconds (43.89 k allocations: 3.605 GiB, 11.14% gc time)
149322456
```

### After
```julia
julia> @time sum(length.(t.Muon_pt))
  2.784472 seconds (21.95 k allocations: 2.032 GiB, 3.90% gc time)
149322456
```

which gets us about a 2x speedup and reduction in allocations for anything that reads iteratively and touches more than 1 basket. 🚀

### DIFF
--- a/test/runtests.jl
+++ b/test/runtests.jl
@@ -522,25 +522,25 @@ end
     @test nmu == 878
 
 
-    nmus = [0]
+    nmus = zeros(Int, Threads.nthreads())
     Threads.@threads for i in 1:length(t)
-        nmus[1] += length(t.Muon_pt[i])
+        nmus[Threads.threadid()] += length(t.Muon_pt[i])
     end
-    @test nmus == [878]
+    @test sum(nmus) == 878
 
 
     @static if VERSION > v"1.3.1"
-        nmus = [0]
+        nmus = zeros(Int, Threads.nthreads())
         Threads.@threads for (i,evt) in enumerate(t)
-            nmus[1] += length(t.Muon_pt[i])
+            nmus[Threads.threadid()] += length(t.Muon_pt[i])
         end
-        @test nmus == [878]
+        @test sum(nmus) == 878
 
-        nmus = [0]
+        nmus = zeros(Int, Threads.nthreads())
         Threads.@threads for evt in t
-            nmus[1] += length(evt.Muon_pt)
+            nmus[Threads.threadid()] += length(evt.Muon_pt)
         end
-        @test nmus == [878]
+        @test sum(nmus) == 878
     end
 
     et = enumerate(t)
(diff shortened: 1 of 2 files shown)


Moelf:
damn... I know I have bugs but I didn't think it would make this much difference, this is amazing

Moelf (approved):

test/runtests.jl
```diff
@@ -459,6 +459,17 @@ end
     rootfile = ROOTFile(joinpath(SAMPLES_DIR, "issue61.root"))
     @test rootfile["Events/Jet_pt"][:] == Vector{Float32}[[], [27.324587, 24.889547, 20.853024], [], [20.33066], [], []]
     close(rootfile)
+
+    # issue 78
+    rootfile = ROOTFile(joinpath(SAMPLES_DIR, "issue61.root"))
+    arr = LazyTree(rootfile,"Events").Jet_pt;
+    _ = length.(arr);
+    if length(arr.buffer) == Threads.nthreads()
```
Moelf, line 467: surely this branch is already determined right?

aminnj: Ah, let me simplify it. I did it at the time because another PR changed the type of the buffer to be a vector with length nthreads().

Moelf:
Some test is locally failing for me. I wonder if `CI` has proper multi threading

aminnj:
You're right, I actually hadn't run with `julia -t xyz runtests.jl`... Let me switch to an atomic integer...

tamasgal:
Ah good catch! Very nice 😄

Moelf:
I wonder if we should check "no overlap" more explicitly. But this looks good for me now.

Moelf (approved):

aminnj:
> I wonder if we should check "no overlap" more explicitly. But this looks good to me now.

I checked one of the example files with sequential numbers and didn't see any discontinuities at basket boundaries. This was true even before the fix here.

aminnj:
OK, just to make sure. I changed 
```julia
br = (ba.fEntry[seek_idx] + 1):(ba.fEntry[seek_idx + 1])
```
to
```julia
br = (ba.fEntry[seek_idx] + 1):(ba.fEntry[seek_idx + 1] + 1)
```
(the other direction of this bug fix) and indeed at least one unit test fails. So if we screw up this kind of overlap indexing, unit tests will definitely complain :)

PR #83, opened 2021-09-06
fix lock again, allow super speed

Moelf:
fix issue created in #82

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,7 +1,7 @@
 name = "UnROOT"
 uuid = "3cd96dde-e98d-4713-81e9-a4a1b0235ce9"
 authors = ["Tamas Gal", "Jerry Ling", "Johannes Schumann", "Nick Amin"]
-version = "0.4.1"
+version = "0.4.2"
 
 [deps]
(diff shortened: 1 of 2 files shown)


Moelf:
```julia
julia> const r = LazyTree(ROOTFile("./Run2012BC_DoubleMuParked_Muons.root"), "Events");

julia> function f3(r)
           h = Hist1D(Int; bins = 1:300)
           Threads.@threads for c in collect(Iterators.partition(1:length(r), 10^5))
               @sync @async for i in c
               r.nMuon[i] !=4 && continue
               r.Muon_charge[i]
               push!(h, maximum(r.Muon_pt[i]))
               end
           end
           h
       end
f3 (generic function with 1 method)

julia> @time f3(r)
  4.401651 seconds (7.87 M allocations: 11.395 GiB, 7.42% gc time, 8.06% compilation time)

edges: 1:300
bin counts: [0, 0, 4257, 13597, 58968, 131841, 163537, 184646, 173327, 155787  …  71, 89, 79, 64, 66, 63, 71, 69, 65, 63]
total count: 5174643

julia> @time f3(r)
  2.990386 seconds (172.05 k allocations: 11.006 GiB, 4.44% gc time)
```

Issue #87, opened 2021-09-08
Uncompressed Streamers cannot be read 


Moelf:
https://github.com/tamasgal/UnROOT.jl/blob/ea2a6ba3244915fe336a13ec3d9ab8b0adb23c38/src/streamers.jl#L109

(on current master branch)
this line throws an `ERROR: Invalid class reference.` when reading a ROOT file that is not compressed at all (even the Streamers are not compressed).

If I print out all the `tag` > 0 from:
https://github.com/tamasgal/UnROOT.jl/blob/ea2a6ba3244915fe336a13ec3d9ab8b0adb23c38/src/streamers.jl#L198-L207

I see that, for a compressed file:
```julia
/home/akako/Downloads/Run2012BC_DoubleMuParked_Muons.root
tag = 0xffffffff
tag = 0xffffffff
tag = 0x8000011e
tag = 0xffffffff
tag = 0x8000011e
tag = 0x80002875
tag = 0x8000011e
tag = 0x80002875
tag = 0x8000011e
tag = 0x80002875
tag = 0x8000011e
tag = 0x800001dd
tag = 0x000001d9
tag = 0x00002871
tag = 0x00005ace
tag = 0x00008d26
tag = 0x0000bf80
tag = 0x0000f1e0
├─ Events
│  ├─ "nMuon"
│  ├─ "Muon_pt"
```

for a bugged uncomressed file:
```
┌ Debug: Unompressed stream at 58521775
└ @ UnROOT ~/.julia/dev/UnROOT/src/streamers.jl:98
┌ Debug: Found 19 streamers, continue with parsing.
└ @ UnROOT ~/.julia/dev/UnROOT/src/streamers.jl:107
tag = 0xffffffff
tag = 0xffffffff
tag = 0xffffffff
tag = 0xffffffff
tag = 0x8000013b
ERROR: Invalid class reference.
Stacktrace:
  [1] error(s::String)
    @ Base ./error.jl:33
```

Moelf:
from 0.1.5:
```julia
julia> ROOTFile("./Run2012BC_DoubleMuParked_Muons_uncompress_1Mevts.root")
ROOTFile("./Run2012BC_DoubleMuParked_Muons_uncompress_1Mevts.root") with 0 entries and 0 streamers.

julia> ROOTFile("./Run2012BC_DoubleMuParked_Muons_uncompress_1Mevts.root")["Events"]
[...]

PR #88, opened 2021-09-09
clean up confusing `origin()` and fix uncompressed reading

Moelf:
fix #87

PR #89, opened 2021-09-09
fix for uncompressed trees

aminnj:
closes https://github.com/tamasgal/UnROOT.jl/issues/87 and supersedes https://github.com/tamasgal/UnROOT.jl/pull/88


tamasgal:
I don't even have a chance 😆

Great, thanks!

aminnj:
So the caveat as this is like a bandaid solution and likely there's a more general one that eventually will/should be found :)

tamasgal:
Step by step `;)`

PR #90, opened 2021-09-10
simplify array creation when not Offsetjagg

aminnj:
When there's no 10 byte offset, VectorOfVectors can use the raw offsets almost directly. Basically all that's needed is
```julia
offset = convert(Vector{Int64}, rawoffsets)
offset .÷= _size
offset .+= 1
```
I got a 0-10% speedup on the uncompressed DoubleMu file with this change. If you know of faster replacement for those 3 lines, we could go even faster! 🚀 :)


aminnj:
@Moelf I should be able to get rid of `offset = convert(Vector{Int64}, rawoffsets)` and use the Int32 offsets due to https://github.com/JuliaArrays/ArraysOfArrays.jl/commit/eca846a8f87e5792c53209f18583b12a732e2b95, but I still get complaints, even though my `ArraysOfArrays` is `v"0.5.4"`...

aminnj:
Thanks to @Moelf for the magic lines to get Int32 offsets everywhere. If there is any speedup, it's slight, but not negative at least.

Moelf:
https://github.com/tamasgal/UnROOT.jl/blob/65bc44d4af4cdc80a94fe7479017d6d46fa5e80b/src/custom.jl#L56

you also need to change this, it's alarming our test doesn't detect this

aminnj:
Test added separately in https://github.com/tamasgal/UnROOT.jl/pull/92

aminnj:
Since
```julia
julia> Tuple{SubArray{UInt8, 1, Vector{UInt8}, Tuple{UnitRange{Int64}}, true}, Vector{Int32}} <: Tuple{Vector{UInt8},Vector{Int32}}
false
```
that means we were incurring the non-trivial cost of `convert()` at the end of `readbasketseek()`. However, we can't just remove the type annotation because then there are downstream issues with the `@view rawdata`, so the `@view` is removed. Total allocations slightly decreases, but performance is basically the same.

PR #93, opened 2021-09-10
small optimization for nonjagged branches

aminnj:
Forgot to show non-jagged code paths some love when we were optimizing jagged branches earlier. This gives a 5-10% improvement in non-jagged branch speed.
```diff
     if J == Nojagg
-        return map(ntoh, reinterpret(T, rawdata))
+        return ntoh.(reinterpret(T, rawdata))
```

### DIFF
--- a/src/root.jl
+++ b/src/root.jl
@@ -201,7 +201,7 @@ function interped_data(rawdata, rawoffsets, ::Type{T}, ::Type{J}) where {T, J<:J
     # the other is where we need to auto detector T bsaed on class name
     # we want the fundamental type as `reinterpret` will create vector
     if J == Nojagg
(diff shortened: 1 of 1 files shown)


Moelf:
```julia
julia> a = [UInt8(1) for _ = 1:3];

julia> map(ntoh, reinterpret(Bool, a))
3-element Vector{Bool}:
 1
 1
 1

julia> ntoh.(reinterpret(Bool, a))
3-element BitVector:
 1
 1
 1
```

aminnj:
I thought the second one is better because BitVector is 8x smaller in memory? Or is it noticeably slower?

Moelf:
I think it throws off our type detection

PR #99, opened 2021-09-12
only show TTrees with the highest cycle number

aminnj:
Since we don't provide a way to get a specific cycle number like `f["Events;2"]` we might as well hide all but the latest ttree. This assumes that TTree TKeys are in order of descending cycle number. We already use this assumption because `f["Events"]` goes to the first fName match. Can someone confirm this assumption is safe?

### before
```julia
julia> UnROOT.samplefile("tree_cycles_hist.root")
ROOTFile with 6 entries and 23 streamers.
/Users/namin/.julia/dev/UnROOT/test/samples/tree_cycles_hist.root
├─ Events (TTree)
│  └─ "Jet_pt"
├─ Events (TTree)
│  └─ "Jet_pt"
├─ myTH1F (TH1F)
├─ myTH1D (TH1D)
├─ myTH2F (TH2F)
└─ myTH2D (TH2D)
```

### after
```julia
julia> UnROOT.samplefile("tree_cycles_hist.root")
ROOTFile with 6 entries and 23 streamers.
/Users/namin/.julia/dev/UnROOT/test/samples/tree_cycles_hist.root
├─ Events (TTree)
│  └─ "Jet_pt"
├─ myTH1F (TH1F)
├─ myTH1D (TH1D)
├─ myTH2F (TH2F)
└─ myTH2D (TH2D)
```


src/displays.jl
```diff
@@ -9,13 +9,19 @@ struct TKeyNode
 end
 function children(f::ROOTFile)
     # display TTrees recursively
+    # subsequent TTrees with duplicate fName will be skipped
+    # since TKey cycle number is guaranteed to be decreasing
     # then all TKeys in the file which are not for a TTree
+    seen = Set{AbstractString}()
```
Moelf, line 15:
```suggestion
    seen = Set{String}()
```

we know `keys()` returns Strings I think

PR #100, opened 2021-09-12
more optimizations

aminnj:
Profiling showed a couple % on separate `materialize` calls for the two offset lines. I guess they weren't being combined? We also were spending a few % in  `full_consistency_checks(A::VectorOfArrays)` . Luckily we can make that a no-op :)

```diff
-        offset .÷= _size
-        offset .+= 1
-        return VectorOfVectors(real_data, offset)
+        offset .= (offset .÷ _size) .+ 1
+        return VectorOfVectors(real_data, offset, ArraysOfArrays.no_consistency_checks)
```

### benchmark

```julia
julia> using UnROOT

julia> const t = LazyTree(ROOTFile("uncompressed_Run2012BC_DoubleMuParked_Muons.root"),"Events");

julia> function foo(t)
           s = 0.0f0
           for evt in t
               s += sum(evt.Muon_pt)
           end
           s
       end
```

### before

```julia
BenchmarkTools.Trial: 15 samples with 1 evaluation.
 Range (min … max):  1.998 s …   2.076 s  ┊ GC (min … max): 5.87% … 5.73%
 Time  (median):     2.018 s              ┊ GC (median):    5.85%
 Time  (mean ± σ):   2.027 s ± 21.783 ms  ┊ GC (mean ± σ):  5.84% ± 0.07%
```

### after

```julia
BenchmarkTools.Trial: 16 samples with 1 evaluation.
 Range (min … max):  1.891 s …   1.940 s  ┊ GC (min … max): 6.22% … 6.12%
 Time  (median):     1.897 s              ┊ GC (median):    6.21%
 Time  (mean ± σ):   1.902 s ± 13.110 ms  ┊ GC (mean ± σ):  6.18% ± 0.06%
```

____

~5% speedup for an uncompressed file reading one branch. Probably means ~2% speedup on the zlib-compressed version.

### DIFF
--- a/src/root.jl
+++ b/src/root.jl
@@ -250,9 +250,8 @@ function interped_data(rawdata, rawoffsets, ::Type{T}, ::Type{J}) where {T, J<:J
             offset = rawoffsets
         end
         real_data = ntoh.(reinterpret(T, rawdata))
-        offset .÷= _size
-        offset .+= 1
-        return VectorOfVectors(real_data, offset)
+        offset .= (offset .÷ _size) .+ 1
+        return VectorOfVectors(real_data, offset, ArraysOfArrays.no_consistency_checks)
     end
 end
 


aminnj:
The majority of `interped_data` is now spent in `real_data = ntoh.(reinterpret(T, rawdata))`.

tamasgal:
I am wondering if we can do it easily in-place...

aminnj:
I played around a bit. It got messy. I wrapped the reinterpreted array (of type `Base.ReinterpretArray{...}`) to get a Vector{T} and then do `ntoh` in place.

before             |  after
:-------------------------:|:-------------------------:
image  |  image

And it works in the REPL with the loop in the original post and gives another 4-5% speedup. But unit tests fail Probably I shouldn't be using `unsafe_*` without knowing the consequences. I even tried some GC protection...
```julia
# real_data = ntoh.(reinterpret(T, rawdata))
n = length(rawdata)÷_size
ptr = pointer(reinterpret(T, rawdata))
real_data = unsafe_wrap(Array, ptr, (n,))
GC.@preserve rawdata real_data begin
    real_data .= ntoh.(real_data)
    vov = VectorOfVectors(real_data, offset, ArraysOfArrays.no_consistency_checks)
end
return vov
```
but no success.

If I run the unit tests by hand in the REPL, they work. If I put printouts right before certain failures, they suddenly work. Seems like GC?

Moelf:
see https://github.com/aminnj/UnROOT.jl/pull/2

aminnj:
I tried that PR but it was the same speed as before (`func1` -- master), so there's clearly something different about the `unsafe_wrap` I did above (`func2`) and yours (`func3`).
```julia
julia> function func1(rawdata)
           T = Int32
           real_data = ntoh.(reinterpret(T, rawdata))
           return real_data
       end

julia> function func2(rawdata)
          T = Int32
          _size = sizeof(T)
          n = length(rawdata)÷_size
          ptr = pointer(reinterpret(T, rawdata))
          real_data = unsafe_wrap(Array, ptr, (n,))
          GC.@preserve rawdata real_data begin
              real_data .= ntoh.(real_data)
           end
           real_data
       end

julia> function func3(rawdata)
           T = Int32
           _eltype = T
           dp = length(rawdata)
           _size = sizeof(T)
           real_data = GC.@preserve rawdata ntoh.(unsafe_wrap(Array, Ptr{_eltype}(pointer(rawdata)), dp÷_size))
           return real_data
       end

julia> @benchmark func1($x) setup=begin x = rand(UInt8,1_000_000) end
 Range (min … max):  164.473 μs …   6.542 ms  ┊ GC (min … max):  0.00% … 89.25%
 Time  (median):     592.193 μs               ┊ GC (median):     0.00%
 Time  (mean ± σ):   602.361 μs ± 682.423 μs  ┊ GC (mean ± σ):  14.76% ± 12.02%

julia> @benchmark func2($x) setup=begin x = rand(UInt8,1_000_000) end
 Range (min … max):  28.773 μs … 222.436 μs  ┊ GC (min … max): 0.00% … 0.00%
 Time  (median):     35.880 μs               ┊ GC (median):    0.00%
 Time  (mean ± σ):   42.340 μs ±  15.488 μs  ┊ GC (mean ± σ):  0.00% ± 0.00%

julia> @benchmark func3($x) setup=begin x = rand(UInt8,1_000_000) end
 Range (min … max):   81.082 μs …   6.492 ms  ┊ GC (min … max):  0.00% … 86.11%
 Time  (median):     470.794 μs               ┊ GC (median):     0.00%
 Time  (mean ± σ):   475.018 μs ± 613.650 μs  ┊ GC (mean ± σ):  14.62% ± 10.77%

```
If we can get `func2` to work safely (so that it works in the REPL and unit tests and everywhere else) that would be super nice!

aminnj:
Should we merge the changes in the original post into `master`? Then we can worry about additional `unsafe_*` optimizations separately, to avoid conflating the two speedups into one.

PR #101, opened 2021-09-13
Faster in-place `ntoh`

Moelf:
You guys will have to tell me how horrible this PR is.

Before:
```julia
julia> using Revise, UnROOT
[ Info: Precompiling UnROOT [3cd96dde-e98d-4713-81e9-a4a1b0235ce9]
cons
julia> const t = LazyTree(ROOTFile("Run2012BC_DoubleMuParked_Muons.root"), "Events", ["Muon_pt"]);

julia> @time  for evt in t
                evt.Muon_pt
              end
  5.021026 seconds (1.93 M allocations: 4.254 GiB, 4.75% gc time, 4.34% compilation time)

julia> @time  for evt in t
                evt.Muon_pt
              end
  4.285786 seconds (58.16 k allocations: 4.159 GiB, 1.68% gc time)

julia> @time  for evt in t
                evt.Muon_pt
              end
  4.255702 seconds (58.16 k allocations: 4.159 GiB, 1.48% gc time)
```

After:
```julia
julia> const t = LazyTree(ROOTFile("Run2012BC_DoubleMuParked_Muons.root"), "Events", ["Muon_pt"]);

julia> @time let x = 0
           for evt in t
               evt.Muon_pt
           end
           x
       end
  4.731603 seconds (1.46 M allocations: 3.674 GiB, 4.55% gc time, 4.94% compilation time)
0

julia> @time  for evt in t
                evt.Muon_pt
              end
  4.096869 seconds (59.20 k allocations: 3.603 GiB, 1.22% gc time)
```


Moelf:
>Won't this mess up the basketarray API?
yep... which is why I closed it.

I think we should disable `array()` API. Maybe only allow if for `raw = true`

aminnj:
Just tested this PR vs `master` on another machine. I get `@btime`s ranging from 1.698 to 1.652 when alternately switching between the two. It seems like there's no time difference. My benchmark is

```julia
julia> using UnROOT

julia> const t = LazyTree(ROOTFile("uncompressed_Run2012BC_DoubleMuParked_Muons.root"), "Events");

julia> function foo(t)
           s = 0.0f0
           for evt in t
               s += sum(evt.Muon_pt)
           end
           s
       end
foo (generic function with 1 method)

julia> @benchmark foo(t)

```

aminnj:
With a benchmark where I don't loop over the elements I get
```julia
julia> function foo2(t)
           for evt in t
               evt.Muon_pt
           end
       end
foo2 (generic function with 1 method)

julia> @benchmark foo2(t) seconds=15 # master
BenchmarkTools.Trial: 17 samples with 1 evaluation.
 Range (min … max):  851.025 ms … 991.843 ms  ┊ GC (min … max): 6.61% … 6.04%
 Time  (median):     876.236 ms               ┊ GC (median):    6.92%
 Time  (mean ± σ):   889.033 ms ±  40.918 ms  ┊ GC (mean ± σ):  6.75% ± 0.29%
 Memory estimate: 1.82 GiB, allocs estimate: 14304.

julia> @benchmark foo2(t) seconds=15 # PR
BenchmarkTools.Trial: 17 samples with 1 evaluation.
 Range (min … max):  769.904 ms … 873.306 ms  ┊ GC (min … max): 5.60% … 5.41%
 Time  (median):     795.748 ms               ┊ GC (median):    5.84%
 Time  (mean ± σ):   800.257 ms ±  23.881 ms  ┊ GC (mean ± σ):  5.78% ± 0.20%
 Memory estimate: 1.26 GiB, allocs estimate: 14602.
```
but it's kind of a dumb benchmark?

Moelf:
at least
>Memory estimate: 1.26 GiB,

src/root.jl
```diff
@@ -200,8 +200,15 @@ function interped_data(rawdata, rawoffsets, ::Type{T}, ::Type{J}) where {T, J<:J
     # the jaggedness comes from having "[]" in TLeaf's title
     # the other is where we need to auto detector T bsaed on class name
     # we want the fundamental type as `reinterpret` will create vector
+    _size = sizeof(eltype(T))
+    dp = length(rawdata)
     if J == Nojagg
-        return ntoh.(reinterpret(T, rawdata))
+        ϖ = convert(Ptr{eltype(T)}, pointer(rawdata))
```
aminnj, line 206: `ϖ` :(

src/root.jl
```diff
@@ -236,26 +242,25 @@ function interped_data(rawdata, rawoffsets, ::Type{T}, ::Type{J}) where {T, J<:J
                 start = rawoffsets[i]+jagg_offset+1
                 stop = rawoffsets[i+1]
                 l = stop-start+1
-                if l > 0
-                    unsafe_copyto!(rawdata, dp+1, rawdata, start, l)
-                    dp += l
-                    offset[i+1] = offset[i] + l
-                else
-                    # when we have an empty [] in jagged basket
-                    offset[i+1] = offset[i]
```
aminnj, line 245: We don't need this `else`?

PR #95, opened 2021-09-10
fix Vector{Bool} branches and factorize `interped_data`

aminnj:
The change in https://github.com/tamasgal/UnROOT.jl/pull/93 results in `BitVector` instead of `Vector{Bool}` which breaks LazyBranch. We didn't have a test for it, but I added one here. I split up the large `interped_data` into smaller chunks for non-, singly-, doubly-jagged branches so that it's easier to specialize on Bool. Evidently this breaks some LV64 test cases which I'm still figuring out.

```julia
TLorentzVector: Error During Test at /Users/namin/.julia/dev/UnROOT/test/runtests.jl:238
  Test threw exception
  Expression: (branch[1]).x == 1.0
  MethodError: interped_data(::Vector{UInt8}, ::Vector{Int32}, ::Type{LorentzVector{Float64}}, ::Type{UnROOT.Nojagg}) is ambiguous. Candidates:
    interped_data(rawdata, rawoffsets, ::Type{T}, ::Type{UnROOT.Nojagg}) where T in UnROOT at /Users/namin/.julia/dev/UnROOT/src/root.jl:196
    interped_data(rawdata, rawoffsets, ::Type{LorentzVector{Float64}}, ::Type{J}) where {T, J<:UnROOT.JaggType} in UnROOT at /Users/namin/.julia/dev/UnROOT/src/custom.jl:76
  Possible fix, define
    interped_data(::Any, ::Any, ::Type{LorentzVector{Float64}}, ::Type{UnROOT.Nojagg})
  Stacktrace:
   [1] basketarray(f::ROOTFile, branch::UnROOT.TBranchElement_10, ithbasket::Int64)
     @ UnROOT ~/.julia/dev/UnROOT/src/iteration.jl:57
```

PR #102, opened 2021-09-15
fix bool branch interpretation

aminnj:
We needed a specialized case to interpret boolean branches as `Vector{Bool}` instead of `BitVector`. I factored out the solution and unit test from https://github.com/tamasgal/UnROOT.jl/pull/95 .

### before
💥🚗 
### after
```julia
julia> using UnROOT

julia> f = UnROOT.samplefile("NanoAODv5_sample.root");

julia> f["Events/L1_ZeroBias"]
1000-element LazyBranch{Bool, UnROOT.Nojagg, Vector{Bool}}:
 1
 1
 1
 1
 1
 1
 1
 ⋮
 1
 1
 1
 1
 1
 1
 1
```

### DIFF
--- a/src/root.jl
+++ b/src/root.jl
@@ -193,6 +193,10 @@ on type `T` and jagg type `J`.
 In order to retrieve data from custom branches, user should defined more speialized
 method of this function with specific `T` and `J`. See `TLorentzVector` example.
 """
+function interped_data(rawdata, rawoffsets, ::Type{T}, ::Type{Nojagg}) where {T<:Bool}
+    # specialized case to get Vector{Bool} instead of BitVector
+    return map(ntoh,reinterpret(T, rawdata))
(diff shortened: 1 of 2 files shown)


src/root.jl
```diff
@@ -193,16 +193,20 @@ on type `T` and jagg type `J`.
 In order to retrieve data from custom branches, user should defined more speialized
 method of this function with specific `T` and `J`. See `TLorentzVector` example.
 """
+function interped_data(rawdata, rawoffsets, ::Type{T}, ::Type{Nojagg}) where {T<:Bool}
```
Moelf, line 196: Bool is a concrete type
```suggestion
function interped_data(rawdata, rawoffsets, ::Type{Bool}, ::Type{Nojagg})
```

PR #104, opened 2021-09-15
Prefix custom structures with _ to hide them

tamasgal:
Since the custom stuff is mostly for test coverage, it makes sense to hide it from the public API. Btw. it also collides with downstream code since the structs are usually defined there and otherwise there is a clash, see https://github.com/tamasgal/UnROOT.jl/pull/95#issuecomment-920161597

### DIFF
--- a/docs/src/advanced/custom_branch.md
+++ b/docs/src/advanced/custom_branch.md
@@ -65,7 +65,7 @@ julia> data, offsets = array(f, "KM3NET_EVENT/KM3NET_EVENT/snapshotHits"; raw=tr
  0x03
    ⋮
    
-julia> UnROOT.splitup(data, offsets, UnROOT.KM3NETDAQHit)
-4-element Vector{Vector{UnROOT.KM3NETDAQHit}}:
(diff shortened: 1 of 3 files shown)


Moelf:
I thought we already have LorentzVector for test coverage

tamasgal:
I'd rather add more tests 😆 I promise I'll do 😉

Moelf:
I mean they could be in the test for sure, I'm saying if you don't need experiment-specific structs live inside UnROOT.jl. I'm just unsure how much you've factored out and how much you still rely on what's inside UnROOT.jl for actual work, so we didn't touch it.

tamasgal:
I think I have factored out everything, so they are really only there for tests. I just realised there is this `DAQHit` where I experimented with an ultra fast in-place-interpretation by defining it as a `primite type` and hacking `Base.getproperty`. That's a huge miss in test coverage but I did not want to forget about it for future implementations of dynamic type and parser generation. I'll probably just remove it for now and create a gist or so...

PR #105, opened 2021-09-15
Increase custom stuff coverage

tamasgal:
This should finally fix the coverage in `custom.jl`


tamasgal:
@Moelf that should be better now `:)`

PR #106, opened 2021-09-15
factorize interpretation

aminnj:
Split up the large interped_data into smaller chunks for non-, singly-, doubly-jagged branches in case we want to cleanly specialize further (like for `Bool`).

Replaces https://github.com/tamasgal/UnROOT.jl/pull/95 .


tamasgal:
OK thanks! I just run some random script, just to make sure everything is fine `;)`

tamasgal:
Alright, it works nicely!

I am just wondering how we should deal with code repetitions in such cases:

```julia
function interped_data(rawdata, rawoffsets, ::Type{Vector{_KM3NETDAQHit}}, ::Type{Nojagg})
    UnROOT.splitup(rawdata, rawoffsets, _KM3NETDAQHit, skipbytes=10)
end
function interped_data(rawdata, rawoffsets, ::Type{Vector{_KM3NETDAQHit}}, ::Type{Offsetjagg})
    UnROOT.splitup(rawdata, rawoffsets, _KM3NETDAQHit, skipbytes=10)
end
```

aminnj:
I had originally tried to make a Union like this
```diff
-function interped_data(rawdata, rawoffsets, ::Type{Vector{_KM3NETDAQHit}}, ::Type{Nojagg})
-    UnROOT.splitup(rawdata, rawoffsets, _KM3NETDAQHit, skipbytes=10)
-end
-function interped_data(rawdata, rawoffsets, ::Type{Vector{_KM3NETDAQHit}}, ::Type{Offsetjagg})
-    UnROOT.splitup(rawdata, rawoffsets, _KM3NETDAQHit, skipbytes=10)
-end
+function interped_data(rawdata, rawoffsets, ::Type{Vector{_KM3NETDAQHit}}, ::Type{Union{Nojagg,Offsetjagg}})
```

but I saw tons of unit test failures like this
```julia
  ArgumentError: cannot reinterpret `UInt8` as `Vector{UnROOT._KM3NETDAQHit}`, type `Vector{UnROOT._KM3NETDAQHit}` is not a bits type
  Stacktrace:
   [1] (::Base.var"#throwbits#241")(S::Type, T::Type, U::Type)
     @ Base ./reinterpretarray.jl:16
   [2] reinterpret(#unused#::Type{Vector{UnROOT._KM3NETDAQHit}}, a::Vector{UInt8})
     @ Base ./reinterpretarray.jl:36
   [3] interped_data(rawdata::Vector{UInt8}, rawoffsets::Vector{Int32}, #unused#::Type{Vector{UnROOT._KM3NETDAQHit}}, #unused#::Type{UnROOT.Nojagg})
     @ UnROOT ~/.julia/dev/UnROOT/src/root.jl:201
```
so the `interped_data` in `root.jl` was called instead of the one in `custom.jl`, presumably because the `JaggType` parameter is more specific in `root.jl`. Any suggestions?

tamasgal:
Hmm... maybe some more fine-graded hierarchy for the `JaggType` could do it, but I kind of struggle to group them in a logical way in the sense that I have no idea what would be the supertype of `Nojagg` and `OffsetJagg`. It's a bit annoying that the `Union` loses the dispatch-battle.

aminnj:
We don't want this right? :)
```julia
for J in (:Nojagg,:Offsetjagg)
    T = _KM3NETDAQHit # can loop over this to get the other struct too
    @eval function UnROOT.interped_data(rawdata, rawoffsets, ::Type{Vector{$T}}, ::Type{$J})
        UnROOT.splitup(rawdata, rawoffsets, $T, skipbytes=10)
    end
 end
```

tamasgal:
I don't know, probably not 😆

Moelf:
maybe it's time to figure out a more sane customization interface

aminnj:
I'm also not opposed to keeping the main `interped_data` as in `master` (as long as constant propagation compiles the right part of the if-else chain, which is hopefully the case...)

PR #110, opened 2021-09-17
add more basic types

aminnj:
1. I took all the types I found on cppreference.com and converted them to the "ROOT types" (13 of them):
`{'short', 'float', 'double', 'int', 'Long64_t', 'char', 'ULong64_t', 'bool', 'unsigned long', 'unsigned short', 'unsigned char', 'long', 'unsigned int'}`
We were missing a couple.

2. Fixed `Vector{Bool}` branch interpretation
3. Took out the `try-catch` in the `LazyBranch` creation in order to get more informative errors since we support most branch types now. Let me know if this is not the case and we can put the safeguard back.


src/root.jl
```diff
@@ -364,9 +364,13 @@ function auto_T_JaggT(f::ROOTFile, branch; customstructs::Dict{String, Type})
                 elseif elname == "unsigned int" 
                     UInt32
                 elseif elname == "unsigned char" 
-                    Char
+                    UInt8
```
Moelf, line 367: what's wrong with `Char`?

aminnj: without that change, I was getting
```julia
julia> ROOTFile("tree_basictypes.root")["t/unsignedchar"]
3-element LazyBranch{Vector{Char}, UnROOT.Offsetjagg, ArraysOfArrays.VectorOfVectors{Char, Vector{Char}, Vector{Int32}, Vector{Tuple{}}}}:
Error showing value of type LazyBranch{Vector{Char}, UnROOT.Offsetjagg, ArraysOfArrays.VectorOfVectors{Char, Vector{Char}, Vector{Int32}, Vector{Tuple{}}}}:
ERROR: ArgumentError: cannot reinterpret `UInt8` as `Vector{Char}`, type `Vector{Char}` is not a bits type
```
and we have
https://github.com/tamasgal/UnROOT.jl/blob/635ed0e46cc72054357a72d52d11aa4eb77a1f2d/src/root.jl#L289-L291

aminnj: after:
```julia
julia> ROOTFile("tree_basictypes.root")["t/char"]
3-element LazyBranch{Vector{Int8}, UnROOT.Offsetjagg, ArraysOfArrays.VectorOfVectors{Int8, Vector{Int8}, Vector{Int32}, Vector{Tuple{}}}}:
 0-element view(::Vector{Int8}, 1:0) with eltype Int8
 Int8[1]
 Int8[1, 1]

julia> ROOTFile("tree_basictypes.root")["t/unsignedchar"]
3-element LazyBranch{Vector{UInt8}, UnROOT.Offsetjagg, ArraysOfArrays.VectorOfVectors{UInt8, Vector{UInt8}, Vector{Int32}, Vector{Tuple{}}}}:
 0-element view(::Vector{UInt8}, 1:0) with eltype UInt8
 UInt8[0x01]
 UInt8[0x01, 0x01]
```

Moelf: that's actually a bug on the reinterpret, reinterpret should becalled as `reinterpret(eltype, Vector{UInt8})`

Moelf:
```julia
Char is a 32-bit AbstractChar type that is the default representation of characters in Julia.
```

nvm, I think we still need `UInt8`

aminnj: Ah right `Char` is fundamentally different (4 bytes vs 1 byte)...
```cpp
root [2] sizeof(char)
(unsigned long) 1
root [3] sizeof(unsigned char)
(unsigned long) 1
root [4] sizeof(int)
(unsigned long) 4
```

PR #112, opened 2021-09-18
[WIP] resurrecting in place `ntoh`

aminnj:
Resurrection of https://github.com/tamasgal/UnROOT.jl/pull/101 (which means some copy-pasting from @Moelf ;) ). Basically, we add a `inplace::Bool` to `interped_data` which toggles between doing `ntoh.(reinterpret(...))` (copy) or an inplace-variant, both returning the same type which means it's still type stable.

But with a small modification, `basketarray()`'s return type depends on the boolean. Probably this can be fixed with a barrier function to discard  the `Ref{UInt8[]}` if called with `inplace=false` (the default.

Unit tests all pass.

```julia
julia> using UnROOT

julia> const t = LazyTree(ROOTFile("uncompressed_Run2012BC_DoubleMuParked_Muons.root"),"Events");

julia> function bar(t)
           for evt in t
               _ = evt.Muon_pt
           end
           nothing
       end
```

### before
```julia
julia> @btime sum(t.nMuon) seconds=10
  204.417 ms (6078 allocations: 469.98 MiB)
0x0000000008e67ad8

julia> @btime bar(t) seconds=10
  712.140 ms (14304 allocations: 1.82 GiB)
```

### after

```julia
julia> @btime sum(t.nMuon) seconds=10
  205.316 ms (6780 allocations: 235.24 MiB)
0x0000000008e67ad8

julia> @btime bar(t) seconds=10
  725.626 ms (16241 allocations: 1.26 GiB)
```

Performance is pretty much the same but with less total allocated memory. If I profile `bar(t)` I see there's a materialization. I thought this should be in place??

image

### DIFF
--- a/src/custom.jl
+++ b/src/custom.jl
@@ -49,7 +49,7 @@ end
 The `interped_data` method specialized for `LorentzVector`. This method will get called by
 [`basketarray`](@ref) instead of the default method for `TLorentzVector` branch.
 """
-function interped_data(rawdata, rawoffsets, ::Type{Vector{LVF64}}, ::Type{Offsetjagg})
+function interped_data(rawdata, rawoffsets, ::Type{Vector{LVF64}}, ::Type{Offsetjagg}, ::Bool)
     _size = 64 # needs to account for 32 bytes header
     dp = 0 # book keeping for copy_to!
     lr = length(rawoffsets)
@@ -73,7 +73,7 @@ function interped_data(rawdata, rawoffsets, ::Type{Vector{LVF64}}, ::Type{Offset
     offset .+= 1
     VectorOfVectors(real_data, offset)
 end
-function interped_data(rawdata, rawoffsets, ::Type{LVF64}, ::Type{J}) where {T, J <: JaggType}
+function interped_data(rawdata, rawoffsets, ::Type{LVF64}, ::Type{J}, ::Bool) where {T, J <: JaggType}
     # even with rawoffsets, we know each TLV is destinied to be 64 bytes
     [
      reinterpret(LVF64, x) for x in Base.Iterators.partition(rawdata, 64)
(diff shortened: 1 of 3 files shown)


Moelf:
I know it probably doesn't get much better but this is very messy :(

Moelf:
actually:
```julia
#before
(14304 allocations: 1.82 GiB)
#after
(16241 allocations: 1.26 GiB)
```

I think this means we shouldn't do this, the number of allocation increases probably caused slow down. And the difference isn't as big as the 2x in the `nMuon` case

aminnj:
I agree it’s not pretty. I’ll keep this an open wip for now since at least it works. If we get inspiration for making it better and truly in place (?) that would be great :)

tamasgal:
I also think that this needs a bit more time `;)` let's keep this floating around...

Moelf:
I wonder if @oschulz has any opinion. In fact, it would be nice to have a variation of `VectorOfVector` that has a field for `Ref` to hold the original byte array...

Moelf:
what happens if you make `Wrapper <:`

https://github.com/JuliaArrays/ArraysOfArrays.jl/blob/393a25de3796ae5dbf69760edcb659dc984ff0f5/src/array_of_similar_arrays.jl#L18

oschulz:
I'm not sure I've understood all the implications here - are you looking for somethings like a `VectorOfVectors` wrapped around an [`UnsafeArray`](https://github.com/JuliaArrays/UnsafeArrays.jl)?

Moelf:
the problem is we have an original byte array `A::Vector{UInt8}`, we want to wrap `VoV` around it except we want to "cast" it to, for example, `B::Vector{Int32}`, but if you cast it by `unsafe_wrap()`, GC thinks `A` is not needed and thus GC-ed and you get bad data in `VoV(B)`.

oschulz:
But why can't we use `ReinterpretArray` if we have to keep the reference to `A` around anyway?

Moelf:
because `ReinterpretArray` is slow for both:
1. `A .= ntoh.(A)`
2. everything user will later on do, such as looping over the branch

oschulz:
> A .= ntoh.(A) (10x slower)

I wonder why ReinterpretArray is performing that much worse, compared to our custom wrappers. Well, probably can't be helped right now.

In that case I would recommend to use a `VectorOfVectors` around a custom wrapper - would that be workable? I don't think creating a special VoV type would increase performance, and it would generate more code to maintain long term.

Moelf:
https://github.com/JuliaLang/julia/issues/42227#issuecomment-1037463738

on Julia master, `y .= ntoh.(y)` is no longer slow given `y` is an `ReinterpretedArray`

aminnj:
It might be that zlib decompression + the calculation is washing out the true benchmark

PR #115, opened 2021-09-26
optimize TBasketKey unpacking

aminnj:
When reading files where the performance is not dominated by interpretation, ~20% of the time is in the `fields...` unpacking. In this file, there are ~175 events per basket, probably because there are many branches.
```julia
julia> t.pass_dxysig
57667-element LazyBranch{Bool, UnROOT.Nojagg, Vector{Bool}}:
 0
 ⋮

julia> UnROOT.numbaskets(t.pass_dxysig.b)
328
```
After fixing the 20% `fields...` unpacking time, 75% of the remaining time is spent in `auto_T_JaggT`. See lower in this thread for the explanation.

**before**
```julia
julia> @benchmark sum(t.pass_dxysig)
BenchmarkTools.Trial: 83 samples with 1 evaluation.
 Range (min … max):  57.313 ms … 70.517 ms  ┊ GC (min … max): 0.00% … 0.00%
 Time  (median):     60.262 ms              ┊ GC (median):    0.00%
 Time  (mean ± σ):   60.802 ms ±  2.639 ms  ┊ GC (mean ± σ):  0.40% ± 1.66%

 Memory estimate: 2.59 MiB, allocs estimate: 27864.
```

**after**
```julia
julia> @benchmark sum(t.pass_dxysig)
BenchmarkTools.Trial: 378 samples with 1 evaluation.
 Range (min … max):  11.502 ms … 20.477 ms  ┊ GC (min … max): 0.00% … 25.99%
 Time  (median):     12.705 ms              ┊ GC (median):    0.00%
 Time  (mean ± σ):   13.239 ms ±  1.600 ms  ┊ GC (mean ± σ):  1.19% ±  4.54%

 Memory estimate: 1.62 MiB, allocs estimate: 16056.
```

### DIFF
--- a/src/types.jl
+++ b/src/types.jl
@@ -59,48 +59,46 @@ end
 
 function unpack(io, T::Type{TBasketKey})
     start = position(io)
-    fields = Dict{Symbol, Union{Integer, String}}()
-    sizehint!(fields, length(fieldnames(T)))
 
-    fields[:fNbytes] = readtype(io, Int32)
-    fields[:fVersion] = readtype(io, Int16)  # FIXME if "complete" it's UInt16 (acc. uproot)
+    fNbytes = readtype(io, Int32)
+    fVersion = readtype(io, Int16)  # FIXME if "complete" it's UInt16 (acc. uproot)
 
-    inttype = fields[:fVersion] <= 1000 ? Int32 : Int64
+    inttype = fVersion <= 1000 ? Int32 : Int64
 
-    fields[:fObjlen] = readtype(io, Int32)
-    fields[:fDatime] = readtype(io, UInt32)
-    fields[:fKeylen] = readtype(io, Int16)
-    fields[:fCycle] = readtype(io, Int16)
-    fields[:fSeekKey] = readtype(io, inttype)
-    fields[:fSeekPdir] = readtype(io, inttype)
-    fields[:fClassName] = readtype(io, String)
-    fields[:fName] = readtype(io, String)
-    fields[:fTitle] = readtype(io, String)
+    fObjlen = readtype(io, Int32)
+    fDatime = readtype(io, UInt32)
+    fKeylen = readtype(io, Int16)
+    fCycle = readtype(io, Int16)
+    fSeekKey = readtype(io, inttype)
+    fSeekPdir = readtype(io, inttype)
+    fClassName = readtype(io, String)
+    fName = readtype(io, String)
+    fTitle = readtype(io, String)
 
     # if complete (which is true for compressed, it seems?)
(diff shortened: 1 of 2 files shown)


Moelf:
>there are ~175 events per basket

this is depressing :(

>probably because there are many branches

isn't each branch basically independent?

aminnj:
I can't remember back to when or where the file was created, but I'm guessing there was some heuristic to decide basket size based on event content (jagged/nonjagged) and number of total branches in the file

aminnj:
If you know how to speed this up, we can get rid of ~100% of the `auto_T_JaggT` time (for this example, at least):
image

aminnj:
Most of the `auto_T_JaggT` is spent in `JaggType` where `branch.fClassName` fails out of the `try-catch` for basic branch types.

https://github.com/tamasgal/UnROOT.jl/blob/db8938009f62f12d994433b81efe5d81eb91708f/src/utils.jl#L64-L70

oddly enough, putting a `!hasproperty(branch, :fClassName) && return Nojagg` before the `try` block eliminates the slowdown.

**before**
```julia
julia> @time for s in propertynames(t); _ = getproperty(t, s)[:]; end
  4.425899 seconds (1.67 M allocations: 242.690 MiB, 0.54% gc time)
```
**after**
```julia
julia> @time for s in propertynames(t); _ = getproperty(t, s)[:]; end
  1.554661 seconds (1.71 M allocations: 245.022 MiB, 1.59% gc time, 6.71% compilation time)
```
(yes dumb example of looping through all branches, but it can happen in the wild ;) )

Moelf (approved):

aminnj:
Another note, the `readtype()`s seem to collectively give a slowdown (but still tiny) around here:
https://github.com/tamasgal/UnROOT.jl/blob/db8938009f62f12d994433b81efe5d81eb91708f/src/types.jl#L70-L84
I found we can make `unpack(..., TBasketKey)` about 40% faster by adding 3 lines after L70 once we know the keylen:
```julia
bytes = read(io, fKeylen - 16)
start = -16
io = IOBuffer(bytes)
```
which accounts for the fact that we already read 16 bytes, then reads the remaining key's bytes in one go and creates a new IOBuffer. The fact that this is faster than the original IOBuffer means that julia saves overhead from asking the disk (cache) for those bytes one by one?

I didn't include it as part of the PR because it might be overkill for the performance it adds.

tamasgal:
That's a totally valid improvement, go for it! :) I recall that I was considering such micro optimisations a few times but hold them back while I was figuring out how things work, just to make it easier to read and adjust the logic. You'll likely find a few other hidden gems like this 😅

Moelf (approved):

Issue #116, opened 2021-09-27
Cannot read `TTree` from file


gipert:
Hi! First of all thank you for providing the package, hope to see it grow and improve in the future!

I just started trying it out and I've found (I think) a bug. I cannot read the `TTree`stored in the attached file [test.root.txt](https://github.com/tamasgal/UnROOT.jl/files/7236227/test.root.txt) (I had to add `.txt` otherwise GitHub does not let me upload it). In particular:
```julia
julia> using UnROOT

julia> f = ROOTFile("test.root")
ROOTFile with 3 entries and 24 streamers.
mage-output.root
├─ InputMacroCommands (TObjString)
└─ NumberOfPrimaries (TParameter<long>)


julia> keys(f)
3-element Vector{String}:
 "InputMacroCommands"
 "NumberOfPrimaries"
 "fTree"

julia> t = LazyTree(f, "fTree")
ERROR: MethodError: no method matching keys(::Nothing)
Closest candidates are:
  keys(::Union{Tables.AbstractColumns, Tables.AbstractRow}) at /home/gipert/.julia/packages/Tables/i6z2B/src/Tables.jl:184
  keys(::Dictionaries.AbstractIndices) at /home/gipert/.julia/packages/Dictionaries/xfzRj/src/AbstractIndices.jl:38
  keys(::OrderedCollections.OrderedSet) at /home/gipert/.julia/packages/OrderedCollections/PRayh/src/ordered_set.jl:95
  ...
Stacktrace:
 [1] LazyTree(f::ROOTFile, s::String)
   @ UnROOT ~/.julia/packages/UnROOT/EoYKp/src/iteration.jl:258
 [2] top-level scope
   @ REPL[18]:1
```
Any idea?

aminnj:
Thanks for the file! Indeed there's something funny going on. For now, you can "solve" it with a bandaid:
```diff
diff --git a/src/streamers.jl b/src/streamers.jl
index 4b14ffb..6ab8144 100644
--- a/src/streamers.jl
+++ b/src/streamers.jl
@@ -589,6 +589,9 @@ function unpack(io, tkey::TKey, refs::Dict{Int32, Any}, ::Type{T}) where {T<:ROO
     mod, typename = split(String(streamer_name), ".")
     streamer = getfield(@__MODULE__, Symbol(typename))
     readfields!(cursor, fields, streamer)
+    if streamer === TBranchElement_10
+        fields[:fIOFeatures] = 0x00
+    end
[...]

PR #117, opened 2021-09-27
default byte for fIOFeatures

aminnj:
Closes https://github.com/tamasgal/UnROOT.jl/issues/116

The `TBranchElement_10` struct expects `fIOfeatures` but `TBranch_12` doesn't read that field (only `TBranch_13` has it):
```cpp
root [11] TBranch::Class()->GetStreamerInfo(12)->ls()
root [12] TBranch::Class()->GetStreamerInfo(13)->ls()
```


aminnj:
Coverage 😊

Issue #114, opened 2021-09-22
Implement `Tables.partitions(t::LazyTree)`


Moelf:
A prototype implementation has been shown at https://gist.github.com/aminnj/79ceeb7ab9026f80e3cc1978124f8370 by @aminnj 

The use case when the analysis is highly columnar (but not single column), users can potentially gain by having SIMD, so imagine computing invariant mass using two columns:
```julia
for table in Tables.partitions(t::LazyTree)
   mass.(table.lep1, table.lep2)
   # or
   @simd for row in table
       compute(row.blah1, row.blah2)
   end
end
```

in this case, the columns of inner `table` has to be contiguous in RAM (then we hope compiler is smart). Actually, @oschulz can you confirm `TypedTable` of `VectorOfVectors` is SIMD-able at all?

Moelf:
In fact, I don't know how to SIMD if a single `row` of `table` is bigger than SIMD size (256/512 bits?)

oschulz:
> @oschulz can you confirm TypedTable of VectorOfVectors is SIMD-able at all?

SIMD for `VectorOfVector`'s columns in a `TypedTables.Table` is not straightforward, because there are two loops involved. But it definitely works for scalar-valued columns (quite an important use case too, after all).

The main advantage of using `VectorOfVector`s will be the ability to use `flatview`, `deepmap`, `adapt(CuArray, table)` and so on.

oschulz:
I should probably also add support for Functors.jl to ArraysOfArrays.jl, to enable `fmap` and `functor`.

Moelf:
>works for scalar-valued columns

yeah but in that case I don't see advantage of `for table in partition(ttree)` because:
1. If you're using scalar column, just do it on the original tree: `ttree.scalar1`
2. If many branches are loaded, the inner `for row in table` will stop SIMD for either VoV or because too many columns

aminnj:
Before I forget to mention it. There are a couple things to note about the above `partitions()` implementation.
1. We should have a more robust way to determine good partitions. I've only ever seen the `fCluster` stuff in NanoAOD. Haven't checked what uproot does, but they have the option for a maximum memory step size in `iterate()`. That's a good idea and calculable from the `TBranch` fields I think.
[...]

PR #118, opened 2021-09-28
basket cluster iteration utilities

aminnj:
Related to https://github.com/tamasgal/UnROOT.jl/issues/114

Adding some utilities to deal with cluster iteration. One can give `clusterranges` either a `LazyTree` or an array of `LazyBranch` and it will return ranges where `fBasketEntry` lines up across all the input branches. That's the most naive thing I could think of given that the `fClusterRangeEnd` is almost always empty. [uproot4](https://github.com/scikit-hep/uproot4/blob/81624d71f0c1c7cefce2db78c069e9f98bea3f65/src/uproot/behaviors/TBranch.py#L1841) does the same thing. Note below that there are 398 baskets, but only 75 clusters.

```julia
julia> using UnROOT ; const t = LazyTree(ROOTFile("Run2012BC_DoubleMuParked_Muons.root"),"Events");

julia> UnROOT.numbaskets(t.nMuon.b)
398

julia> UnROOT._clusterranges([t.nMuon, t.Muon_pt])
75-element Vector{UnitRange{Int64}}:
 1:821695
 821696:1643390
 ⋮
 59983736:60805430
 60805431:61540413

julia> UnROOT._clusterranges(t)
75-element Vector{UnitRange{Int64}}:
 1:821695
 821696:1643390
 ⋮
 59983736:60805430
 60805431:61540413
```
The idea would be to then have
```julia
# until the @async issue is fixed in `Arrow.write()`
function _lockedget(t::LazyTree, r::UnitRange)
    f = getproperty(t,first(propertynames(t))).f
    lock(f)
    try
        return t[r]
    catch
    finally
        unlock(f)
    end
end
Tables.partitions(t::LazyTree) = (_lockedget(t, r) for r in _clusterranges(t))
```

There's also a function to get the number of bytes per cluster. The logic is there and can be modified based on what we want to actually _do_ with this information (combine clusters until we reach `x` MB?). Note that the cluster sizes are around 30 MB. IIrc, there's some 30MB default AutoFlush parameter for TTrees somewhere, so that's a promising sign.

```julia
julia> UnROOT._clusterbytes(t; compressed=true) ./ 1024^2
75-element Vector{Float64}:
 28.633505821228027
 27.9722261428833
  ⋮
 29.023967742919922
 25.91922950744629
```

For reading into memory instead of piping back to disk, `compressed=false` may be a better metric.

```julia
julia> UnROOT._clusterbytes(t; compressed=false) ./ 1024^2
75-element Vector{Float64}:
 58.74042271086661
 55.89151628176583
  ⋮
 57.50102683642917
 51.386066935034215
```


tamasgal:
Rampage 😆

Looks good to me!

tamasgal (approved):

Moelf:
it turns out that `common_entry_offsets` is not used anymore and uproot4 just don't care where baskets are:
1. when iterating over chunks, if the next chunk happens to include some of the already-read baskets, they will re-use it
2. other wise they cut off baskets wherever (according to default or user-specified chunk size).

Issue #120, opened 2021-09-30
Better/safer printout


aminnj:
Sometimes I accidentally type `print(t)` in a script and then have to kill my terminal tab as millions of rows flood into stdout. 

We get a nice compact representation with the "default repr" in the REPL
```julia
julia> p
 Row │ nMuon   Muon_pt
     │ UInt32  SubArray{Float3
─────┼─────────────────────────
 1   │ 0       []
 2   │ 2       [19.9, 15.3]
 3   │ 0       []
 4   │ 0       []
 5   │ 0       []
 6   │ 0       []
 7   │ 2       [22.2, 4.43]
 8   │ 0       []
 9   │ 0       []
 10  │ 0       []
  ⋮  │   ⋮            ⋮
                90 rows omitted
```
which you can reproduce with `display(p)` or `println(repr(p, context=:limit=>true))`, the former of which is unintuitive for a newcomer and the latter unintuitive unless you spend 10 mins tracing through `show/display/repr` in Base.

I think pandas has a sane approach here:
```python
>>> import pandas as pd
>>> df = pd.DataFrame(range(1000))
>>> df
       0
0      0
1      1
2      2
3      3
4      4
..   ...
995  995
996  996
997  997
998  998
999  999

[1000 rows x 1 columns]
>>> print(df)
       0
0      0
1      1
2      2
3      3
4      4
..   ...
995  995
996  996
997  997
998  998
999  999

[1000 rows x 1 columns]
```
so, I propose that we match this behavior (always give compact output).

tamasgal:
OK, I have mixed feelings about it 😅 I was always a bit annoyed that pandas skips rows when I do `print` because I usually would assume that `print` prints everything (in the old-sense of printing something on a physical printer, where you expect the whole thing and not skipping the middle part).

On the other hand, `numpy` does the same, so I understand the consistency issue. Maybe they started it and other devs just adopted it? Also, we don't have much fine-graded options for the "printer" like "from page 4-10", so my argument above is lacking a good foundation 😆

Moelf:
>assume that print prints everything

I agree.

>for the "printer" like "from page 4-10"

I say we add this and other issues and one more for https://github.com/ronisbr/TerminalPager.jl as hacktoberfest 2021 ;)

aminnj:
[...]

PR #121, opened 2021-09-30
match DataFrames behavior for `show`

aminnj:
Closes https://github.com/tamasgal/UnROOT.jl/issues/120

This PR makes `show` behave like `DataFrames`, so the table in the above issue is identical, namely

`LazyTree`/`DataFrame` | REPL | script
-- | -- | --
`t`/`df` | short | N/A
`show` | short | short
`println` | full | full
`display` | short | full

The change here is essentially a minimal version of 
https://github.com/JuliaData/DataFrames.jl/blob/7fcb93baf95b985f1ca4a691eb066c7f5f72b3f4/src/abstractdataframe/show.jl#L357-L367
which calls a custom show with kws for eliding rows for `stdout`.


Moelf:
is this test worthy....?

tamasgal:
I think this looks fine `;)`

tamasgal (approved):

PR #129, opened 2021-10-05
test zstd compression

aminnj:


aminnj:
Does anyone actually use zstd??  It was supposed to be a compromise between zlib and lz4. I feel like there was hype around it but then people settled back into zlib/lzma/lz4.

Moelf:
>Does anyone actually use zstd?

we should have some debugging utility to quickly see information about a `.root` file

Moelf:
https://indico.cern.ch/event/695984/contributions/2872933/attachments/1590457/2516802/ZSTD_and_ZLIB_Updates_-_January_20186.pdf#page=8

I almost forgot which one sucked the most, now I remembered, `lzma` sucks, zstd is pretty fine but as good as lz4 all things considered.

Issue #130, opened 2021-10-06
LazyTree printout in a notebook


aminnj:
At some point, the pretty table in a notebook got broken (it thinks we're in a terminal)...
![image](https://user-images.githubusercontent.com/5760027/136297806-0a00fc68-fcd8-4a0f-868f-7bd74311978b.png)

tamasgal:
Yep, that should be it!

PR #132, opened 2021-10-07
custom html repr

aminnj:
Closes https://github.com/tamasgal/UnROOT.jl/issues/130

We have a little bit of mix-and-match between PrettyTables and DataFrames...

(top is our representation using PrettyTables+DF style, bottom is pure DF)
image


aminnj:
We can asymptotically approach 100% coverage with a nice little green badge in the README...we just need to add a ton more trivial code and cover it. ;)

PR #140, opened 2021-12-07
Add `view` of LazyTree

Moelf:
if `tf` is a `LazyTree`:
1. `tf[scalar]` gives you `LazyEvent`
2. `tf[AbstractVector]` is always eager, will materialize the events (making disk I/O)
3. `@view tf[AbstractVector]` is lazy <--- this PR makes this work

before this PR, `@view` would give a` SubArray` with `eltype == LazyEvent`, which isn't ideal, now it gives you a proper `LazyTree` table still, with minimal allocation thanks to `view`


Moelf:
ehhh, tests pass locally for me, not sure why allocation grows...

Issue #142, opened 2022-01-20
Test suite requires more than one thread


aminnj:
I'll act as a human CI for an M1 mac ;)

This multi threading unit test fails for me when I use the julia default of 1 thread. However everything succeeds if I use several threads (eg `JULIA_NUM_THREADS=4  julia runtests.jl`)

```julia
$ julia runtests.jl
Test Summary:                    | Pass  Total
Parallel and enumerate interface |    9      9
Multi threading: Test Failed at /Users/namin/.julia/dev/UnROOT/test/runtests.jl:658
  Expression: count((>)(0), nmus) > 1
   Evaluated: 1 > 1
Stacktrace:
 [1] macro expansion
   @ /Applications/Julia-1.7.app/Contents/Resources/julia/share/julia/stdlib/v1.7/Test/src/Test.jl:445 [inlined]
 [2] macro expansion
   @ ~/.julia/dev/UnROOT/test/runtests.jl:658 [inlined]
 [3] macro expansion
   @ /Applications/Julia-1.7.app/Contents/Resources/julia/share/julia/stdlib/v1.7/Test/src/Test.jl:1283 [inlined]
 [4] top-level scope
   @ ~/.julia/dev/UnROOT/test/runtests.jl:648
Multi threading: Test Failed at /Users/namin/.julia/dev/UnROOT/test/runtests.jl:666
  Expression: count((>)(0), nmus) > 1
   Evaluated: 1 > 1
Stacktrace:
 [1] macro expansion
   @ /Applications/Julia-1.7.app/Contents/Resources/julia/share/julia/stdlib/v1.7/Test/src/Test.jl:445 [inlined]
 [2] macro expansion
   @ ~/.julia/dev/UnROOT/test/runtests.jl:666 [inlined]
 [3] macro expansion
   @ /Applications/Julia-1.7.app/Contents/Resources/julia/share/julia/stdlib/v1.7/Test/src/Test.jl:1283 [inlined]
 [4] top-level scope
   @ ~/.julia/dev/UnROOT/test/runtests.jl:648
Test Summary:   | Pass  Fail  Total
Multi threading |    7     2      9
ERROR: LoadError: Some tests did not pass: 7 passed, 2 failed, 0 errored, 0 broken.
in expression starting at /Users/namin/.julia/dev/UnROOT/test/runtests.jl:645
```

aminnj:
Ah wait sorry, that test will always fail with only 1 thread :)....
```julia
[...]

PR #143, opened 2022-01-21
fix unit tests failing when run with 1 thread

aminnj:
Fixes https://github.com/tamasgal/UnROOT.jl/issues/142


tamasgal:
Looks good to me! Maybe we should indicate that these tests are skipped, so that devs know they have to check those too...

PR #161, opened 2022-04-07
add TLeafS

### LINKED ISSUE #160: LazyTree construction errors : UndefVarError: TLeafS / UndefVarError: TBasket not defined
Hello, first of all thanks for this amazing package and the fast reaction times (I've started looking into it on monday and experienced the compling error as my first obstacle)!

I've managed to read in the sample file in the quick start sections and compare that data / a generated histogram with one generated from ROOT / with the TBrowser.

Now I'm trying to do that with .root files from ATLAS OpenData (since they have the notebooks for comparison), but for the files I've tried it reads the files but doesn't display it in a "tree-like fashion" though I'm not sure if that limited / just not done for files too large for example.
My main issue is that I'm getting an error when trying to contruct a LazyTree.

For my first case (as far as I understand it "grabs" the whole tree, while ftree=LazyTree(f, "mini", ["lep_n"]) is the specific branch)

```
Using UnROOT

f=ROOTFile("http://opendata.atlas.cern/release/samples/MC/mc_105987.WZ.root")
ftree=LazyTree(f, "mini")

```
I get "UndefVarError: TLeafS not defined" .

For two others (https://atlas-opendata.web.cern.ch/atlas-opendata/samples/2020/GamGam/Data/data_A.GamGam.root and https://atlas-opendata.web.cern.ch/atlas-opendata/samples/2020/1largeRjet1lep/MC/mc_361106.Zee.1largeRjet1lep.root )
I get "UndefVarError: TBasket not defined" in that case.

I've found [this issue](https://github.com/JuliaHEP/UnROOT.jl/issues/14) while looking through the github.
That said I'm not sure if it's a potential bug, a file issue or an user issue (since I'm new to Root and Julia :) ).

Moelf:
partially fix #160

### DIFF
--- a/src/bootstrap.jl
+++ b/src/bootstrap.jl
@@ -203,6 +203,41 @@ end
 primitivetype(l::TLeafI) = l.fIsUnsigned ? UInt32 : Int32
 
 # FIXME this should be generated and inherited from TLeaf
+@with_kw struct TLeafS
+    # from TNamed
+    fName
+    fTitle
+
+    # from TLeaf
+    fLen
+    fLenType
+    fOffset
+    fIsRange
+    fIsUnsigned
+    fLeafCount
+
+    # own fields
+    fMinimum
+    fMaximum
+end
+
+function parsefields!(io, fields, T::Type{TLeafS})
+    preamble = Preamble(io, T)
+    parsefields!(io, fields, TLeaf)
+    fields[:fMinimum] = readtype(io, Int16)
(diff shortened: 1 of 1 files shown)


tamasgal:
Ah that was quick 😆 I was already looking. Now I am investigating this `TBasket` thing...

Moelf:
yeah as shown in the other thread, we can already unpack it and we should. except I don't know what's the relationship between that basket to the branch user wants to read

Issue #165, opened 2022-04-25
Problem with reading branches of C-style arrays


briederer:
I just tried to read one of my Root-files which contains a TTree with TBranches of the following types
  - primitive types: i.e. integer and double branches
  - array types: i.e. a C-Style array of doubles like `double mat[9]`

In principle it seems to be related to my old issue #9 so I decided to dig a bit deeper with the current version.

Loading my file with `ROOTFile()` works without a problem. However when trying to create the `LazyTree` of the said TTree I get the error message:
```julia
Error displaying LazyTree with 6 branches:
n, trsigma2, sigma, sigma2, sigma3...
: ArgumentError: invalid Array dimensions
```
The stacktrace lead me to the `interped_data(rawdata, rawoffsets, ::Type{T}, ::Type{J}) where {T, J<:JaggType}` function. 
It turns out that my data seems to be a `NooffsetJag` JagType which seems kind of reasonable to me. Also reinterpreting works without a problem and indeed the variable `real_data` contains the real data stored in the Root-File.
So in fact the problem only occurs when passing the real_data and the offset to `VectorOfVectors`. The problem seems to be the following: In my case the offset-vector is simply an empty vector `Int32[]` which causes the VectorOfVectors to call the function `similar` to init a vector of size `-1`.

So to sum it up, it seems that UnROOT is able to read my data (and I guess in turn also the file from #9) but simply the wrapping of the data fails. I guess it should be possible to fix it by changing the part for `NooffsetJag` in the [code](https://github.com/JuliaHEP/UnROOT.jl/blob/81b6ca90b722336342f9d60d9ab1971cb4404fe6/src/root.jl#L266-L268) and catch cases when rawoffsets is empty. But I don't know whether this would create other problems.

I really hope we can figure this out, because being able to finally use UnROOT for my analysis would make my analysis much more efficient.

briederer:
Follow-up:
Simply replacing the lines I've mentioned in the previous post by:
```julia
...
else
    #offset = rawoffsets # not needed since set implicitly in return
    real_data = ntoh.(reinterpret(T, rawdata))
    return VectorOfVectors(real_data, Int32[1])
end
```
gives kind of a weird behavior for the lazy tree.
```julia
file = ROOTFile("path/to/file")
lt = LazyTree(file,"tree",["branch"]) # branch contains 30 entries with each arrays of the Root-type `[9]/D`
```
Outputs:
```julia
julia> lt[1] 
[...]

PR #167, opened 2022-05-05
Reading of c-arrays + more customstructs

### LINKED ISSUE #9: Reading structs into higher-dimensional Arrays
ROOT offers different ways to store arrays or even structs in `TBranches`. I added a small ROOT-File ([test_array.root.zip](https://github.com/tamasgal/UnROOT.jl/files/5108576/test_array.root.zip)) which contains a `TTree` called 'arrays' which contains a scalar, a 6d-vector and a 2x3-matrix and another `TTree` called 'structs' which contains again the same scalar as 'arrays' and a struct which resembles again a 2x3-matrix but stored as a 6d-vec and is defined in the following way
`struct m23 {double el[6];};`

So the file was created in the following way (may be useful when trying to retrieve data in Julia):
```
struct m23 {double el[6];};

m23 mat_23 = {1.0,2.0,3.0,4.0,5.0,6.0};
int n = 1;
double vec[6] = {1.0,2.0,3.0,4.0,5.0,6.0};
double mat[2][3] = {{1.0,2.0,3.0},{4.0,5.0,6.0}};

TFile *file = new TFile("test_array.root","Recreate");
TTree *tree_arr = new TTree("arrays","tree filled with arrays")
TTree *tree_str = new TTree("structs","tree filled with struct");

tree_arr -> Branch("nInt",&n);
tree_arr -> Branch("6dVec",&vec,"[6]/D");
tree_arr -> Branch("2x3Mat",&mat,"[2][3]/D");
tree_arr -> Fill();

tree_str -> Branch("nInt",&n);
tree_str -> Branch("2x3mat",&mat,"el[6]/D");
tree_str -> Fill();

file -> Write("",TObject::kOverwrite);
```

In the ROOT command-line the `TTrees` look like this in the end
```
root [1] arrays -> Scan();
***********************************************************
*    Row   * Instance *      nInt * 6dVec.6dV * 2x3Mat.2x *
***********************************************************
*        0 *        0 *         1 *         1 *         1 *
*        0 *        1 *         1 *         2 *         2 *
*        0 *        2 *         1 *         3 *         3 *
*        0 *        3 *         1 *         4 *         4 *
*        0 *        4 *         1 *         5 *         5 *
*        0 *        5 *         1 *         6 *         6 *
***********************************************************
root [2] structs -> Scan();
***********************************************
*    Row   * Instance *      nInt * 2x3mat.2x *
***********************************************
*        0 *        0 *         1 *         1 *
*        0 *        1 *         1 *         2 *
*        0 *        2 *         1 *         3 *
*        0 *        3 *         1 *         4 *
*        0 *        4 *         1 *         5 *
*        0 *        5 *         1 *         6 *
***********************************************
```

When I try to read this in Julia using the UnROOT package I get the following
```
julia> file=ROOTFile("/hosts/nashome/riederer_bernd/test_array.root")
ROOTFile("/hosts/nashome/riederer_bernd/test_array.root") with 2 entries and 17 streamers.

julia> keys(file)
2-element Array{String,1}:
 "arrays" 
 "structs"

julia> array(file,"arrays")
ERROR: Branches with multiple leaves are not supported yet.
Stacktrace:
 [1] #array#120(::Bool, ::typeof(array), ::ROOTFile, ::String) at ~/.juliapro/JuliaPro_v1.2.0-1/packages/UnROOT/KN5MU/src/root.jl:153
 [2] array(::ROOTFile, ::String) at ~/.juliapro/JuliaPro_v1.2.0-1/packages/UnROOT/KN5MU/src/root.jl:139
 [3] top-level scope at none:0

julia> array(file,"structs")
ERROR: Branches with multiple leaves are not supported yet.
Stacktrace:
 [1] #array#120(::Bool, ::typeof(array), ::ROOTFile, ::String) at ~/.juliapro/JuliaPro_v1.2.0-1/packages/UnROOT/KN5MU/src/root.jl:153
 [2] array(::ROOTFile, ::String) at ~/.juliapro/JuliaPro_v1.2.0-1/packages/UnROOT/KN5MU/src/root.jl:139
 [3] top-level scope at none:0

julia> DataFrame(file,"arrays")
1×3 DataFrame
│ Row │ nInt  │ 6dVec   │ 2x3Mat  │
│     │ Int32 │ Float64 │ Float64 │
├─────┼───────┼─────────┼─────────┤
│ 1   │ 1     │ 1.0     │ 1.0     │

julia> DataFrame(file,"structs")
1×2 DataFrame
│ Row │ nInt  │ 2x3mat  │
│     │ Int32 │ Float64 │
├─────┼───────┼─────────┤
│ 1   │ 1     │ 1.0     │

```

It seems that, while `array` is refusing to handle multidimensional objects in a `TBranch`, the function `DataFrame` is able to get the scalar and the first entry of each multidimensional object. 
I tested this with another file and `DataFrame` returns correctly all Rows of scalar-entries and next to it in each case the first element of the higher-dimensional object. So it looks like this:
```
julia> DataFrame(file,"structs")
1×5 DataFrame
│ Row │ nInt  │  2x3mat   │
│     │ Int32 │  Float64  │
├─────┼───────┼───────────┤
│ 1   │ 1     │ mat.el[0] │
│ 2   │ 2     │ mat.el[0] │
│ 3   │ 3     │ mat.el[0] │
│ 4   │ 4     │ mat.el[0] │

```

I would like to request to extend the support for multidimensional objects at least for the `array` function.

briederer:
After prior discussion in #165 and #166 respectively I decided to put a little bit of effort into extending it to fix #9 finally.

For that I simply enhanced the customstruct to an arbitrary dimensional `FixSizeArray` and parse the fLeaf.fTitle for the decomposition. Everything else works as expected.

Two caveats:
  1. Since all of it relies on wrapping the `StaticArrays.SArray` struct nicely we are also limited to there specifications. With the current implementation of the struct `Base.sizeof()` does not work out of the box due to a missing 4.th parameter in the SArray-type. (see: https://github.com/briederer/UnROOT.jl/blob/3785a042a9cd9855e5f0b9ee18e3b89448bb15d3/src/custom.jl#L49-L53). This means we either add an (for this purpose) unnecessary parameter to the struct or we simply overwrite the `Base.sizeof` method as is done here.
  2. I've also added (like it is usually done in Julia) aliases for `FixSizeVector` and `FixSizeMatrix` to improve readability in the output. However due to some restrictions (see  JuliaLang/julia#40448) the aliases would only be seen for the User if the additional types are exported. But this is not my choice :wink:

Closes #9

### DIFF
--- a/src/UnROOT.jl
+++ b/src/UnROOT.jl
@@ -3,6 +3,7 @@ module UnROOT
 using LazyArrays
 import Mmap: mmap
 export ROOTFile, LazyBranch, LazyTree
+#export FixSizeArray, FixSizeMatrix, FixSizeVector  # would make it more readable in printed tables but not necessary
 
 import Base: close, keys, get, getindex, getproperty, show, length, iterate, position, ntoh, reinterpret
 ntoh(b::Bool) = b
--- a/src/custom.jl
+++ b/src/custom.jl
@@ -31,23 +31,33 @@ end
 # Custom struct interpretation
 abstract type CustomROOTStruct end
 
-struct FixLenVector{N, T} <: AbstractVector{T}
-    vec::SVector{N, T}
+# Fixed-size c-style array
+struct FixSizeArray{S <: Tuple, T, N} <: AbstractArray{T, N}
+    arr::SArray{S, T, N}
 end
-(::Type{FixLenVector{N, T}})() where {N, T} = FixLenVector(zero(SVector{N, T}))
-Base.length(x::FixLenVector) = length(x.vec)
-Base.length(::Type{FixLenVector{N, T}}) where {N, T} = N
-Base.size(x::FixLenVector) = size(x.vec)
-Base.eltype(x::FixLenVector) = eltype(x.vec)
-Base.iterate(x::FixLenVector) = iterate(x.vec)
-Base.iterate(x::FixLenVector, n) = iterate(x.vec, n)
-Base.getindex(x::FixLenVector, n) = getindex(x.vec, n)
-function Base.reinterpret(::Type{FixLenVector{N, T}}, v::AbstractVector{UInt8}) where {N, T}
+const FixSizeVector{N, T} = FixSizeArray{Tuple{N}, T, 1}
+const FixSizeMatrix{M, N, T} = FixSizeArray{Tuple{M, N}, T, 2}
+
+(::Type{FixSizeArray{S, T, N}})() where {S, T, N} = FixSizeArray(zero(SArray{S, T, N}))
+Base.length(x::FixSizeArray) = length(x.arr)
+Base.length(t::Type{<:FixSizeArray}) = length(fieldtype(t,:arr))
+Base.size(x::FixSizeArray) = size(x.arr)
+Base.eltype(x::FixSizeArray) = eltype(x.arr)
+Base.iterate(x::FixSizeArray) = iterate(x.arr)
+Base.iterate(x::FixSizeArray, n) = iterate(x.arr, n)
+Base.getindex(x::FixSizeArray, n...) = getindex(x.arr, n...)
+# need this workaround since Base.sizeof is only correctly working for SArray
+# with all 4 template arguments (note only 3 are used in FixSizeArray struct).
+# Could get rid off this if it would be possible to defer the total size in our
+# wrapper struct i.e. see https://github.com/JuliaLang/julia/issues/18466
+Base.sizeof(::Type{<:FixSizeArray{S,T,N}}) where {S, T, N} =  sizeof(T)*prod(fieldtypes(S))
+function Base.reinterpret(::Type{<:FixSizeArray{S,T,N}}, v::AbstractVector{UInt8})  where {S, T, N}
     vs = reinterpret(T, v)
     @. vs = ntoh(vs)
-    FixLenVector(SVector{N, T}(vs))
+    FixSizeArray(SArray{S, T, N}(vs))
 end
-function interped_data(rawdata, rawoffsets, ::Type{T}, ::Type{Nojagg}) where {T <: FixLenVector}
+
+function interped_data(rawdata, rawoffsets, ::Type{T}, ::Type{Nojagg}) where {T <: FixSizeArray}
     n = sizeof(T)
     [
      reinterpret(T, x) for x in Base.Iterators.partition(rawdata, n)
--- a/src/iteration.jl
+++ b/src/iteration.jl
@@ -23,9 +23,9 @@ end
 
 function array(f::ROOTFile, branch; raw=false)
     ismissing(branch) && error("No branch found at $path")
-    # (!raw && length(branch.fLeaves.elements) > 1) && error(
-    #     "Branches with multiple leaves are not supported yet. Try reading with `array(...; raw=true)`.",
-    # )
+    !raw && (!haskey(f.customstructs, branch.fName) && length(branch.fLeaves.elements) > 1) && error(
+        "Branches with multiple leaves are not supported yet. Try reading with `array(...; raw=true)` or by including this branch into `customstructs`.",
+    )
 
     rawdata, rawoffsets = readbranchraw(f, branch)
     if raw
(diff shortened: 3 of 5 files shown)


Moelf:
```
julia> a = @SMatrix [1 2 3; 4 5 6]
2×3 SMatrix{2, 3, Int64, 6} with indices SOneTo(2)×SOneTo(3):
 1  2  3
 4  5  6

julia> sizeof(a)
48
```

it should work?

Moelf:
I forgot why we can't use `SArray` directly actually, I read the last discussion, it seems to be because we build the `buffer` by calling the constructor:
https://github.com/JuliaHEP/UnROOT.jl/blob/7d3712202b0c9bc975937dac5258e20321c3659d/src/iteration.jl#L116

and of course the StaticArray doesn't have constructor. But we can work around this by defining out own:
```julia
_initialize(::Type{T}) = T()
_initialize(::Type{T<:SArray}) = zero(T)
```

briederer:
Indeed in this case it works. But this is due to the fact, that the actual type of your SMatrix is `SMatrix{2, 3, Int64, 6}` which is an alias for `SArray{Tuple{2, 3}, Int64, 2, 6}`.
However the inner type in "my" `FixSizeArray` is missing the last entry in the type arguments (i.e. it is `SArray{Tuple{...}, T, 2}` which is basically a subtype since the last argument can be deduced from the product of all `Tuple{...}` entries. So in fact the wrapped struct is not exactly like in your example.
However it could be adapted by adding everywhere this 4th argument, which seems unnecessary to me though.

briederer:
> I forgot why we can't use SArray directly actually, I read the last discussion, it seems to be because we build the buffer by calling the constructor:
> https://github.com/JuliaHEP/UnROOT.jl/blob/7d3712202b0c9bc975937dac5258e20321c3659d/src/iteration.jl#L116
> and of course the StaticArray doesn't have constructor. But we can work around this by defining our own:
> ```julia
> _initialize(::Type{T}) = T()
> _initialize(::Type{T<:SArray}) = zero(T)
> ```
> this way we don't have to wrap them?

this way we don't have to wrap them?
I am not sure this would work?
First problem is that according to the last discussion the arrays are `Nojagg` and thus would not even enter this part of the code. However even if you circumvent this problem I am not sure if calling a `VectorOfVectors(StaticArrays(),...)` wouldn't be a bit of an overkill (too many nested vector-like structs).
But I agree that getting rid off the wrapper would be a good thing.

briederer:
I've added now also the possibility to allwo the customstructs dict to allow for special types simply by branchnames.

This enables a bit more customizability if someone needs special interpretation of some branches.
E.g. i've got one branch called `complex[2]/D` containing complex numbers stored as two entries. Since it is not a C++-class the branch has no `fClassName` field and the usual approach does not work. Therefore searching for names too is a possible workaround.

briederer:
To be honest I just realised that the last addition is everything that I need for any of my "special branches".

So if you think that the other parts of this PR (and what has been in added in #166) adds too much overhead/unnecessary stuff I would be totally fine with removing all of that. As long as defining "customstructs" also via fTitle remains I am happy 😊

briederer:
Btw. why are tests failing?
I can't access the CI-logs so I don't now what's happening there. Will check if tests fail locally too, tomorrow.

Moelf:
https://github.com/JuliaHEP/UnROOT.jl/runs/6309779626?check_suite_focus=true

is log somehow not visible?

Moelf:
does test pass for you locally? `]test UnROOT`

briederer:
Well they are accessible but it says "The log was not found" 🤔
image

briederer:
> does test pass for you locally? `]test UnROOT`

Will try it tomorrow.

briederer:
Tried now the tests locally. The failing one is this:
https://github.com/JuliaHEP/UnROOT.jl/blob/7d3712202b0c9bc975937dac5258e20321c3659d/test/runtests.jl#L733-L738

It seems that this file has a `TBranch` called `"Type"` populated with `TLeafC` which is a variable size string. Problem that occurs is that this  has a `leaf.fLen > 1` and thus enters the condition added in #166. Before my current modification (i.e. extending to arbitrary size arrays) the test did work because it simply reinterpreted the string wrong (as a `FixLenVector` instead of a string) which is not covered by the tests.

Fixed it by simply leaving it as `FixSizeVector{leaf.fLen,UInt8}` which is a "untranslated" vector of chars. I didn't manage to adopt the output to produce a string for each event (problem with no definite size-strings when reinterpreting) but I guess this could be adapted.

Moelf:
this is partly addressed in https://github.com/JuliaHEP/UnROOT.jl/pull/188

briederer:
Sadly it appears I'll never find time to finish this and (probably) won't need it any time in the near future.
Therefore I am closing the PR now.
In any case someone wants to pick it up again, feel free. I will not delete the branch.

PR #181, opened 2022-09-07
Drop TypedTables as dependency

Moelf:
one could think of this as:
1. implement the basic strategy of TypedTables in-house (i.e. wrapping `NamedTuple` as Table)
2. but provide indexing, slicing, and view according to our need
3. we should be compatible with Tables.jl interface, so user can call `DataFrame, TypedTables.Table` on the `LazyTree`

TODO:
- [ ] @tamasgal test if this still works with your work flow in case you were using some internal functions?
- [x] I will Benchmark @Moelf


oschulz:
We might also utilize [StructArrays.jl](https://github.com/JuliaArrays/StructArrays.jl), it has recently become more lightweight, dependency-wise.

Moelf:
our array (the branches) are not in memory, I doubt StructArrays would help

oschulz:
> our array (the branches) are not in memory

I know - but if each column implements the `AbstractArray` interface, we might just be able to put the on-disk columns directly into a StructArray. We're currently [exploring this approach](https://github.com/legend-exp/LegendHDF5IO.jl/pull/18) for HDF5-back on-disk data wrapped into arrays (with TypedTables, but should work with StructArrays as well).

Moelf:
```julia
[ Info: 1st run
main(t): 15.572835 seconds (20.07 M allocations: 16.459 GiB, 2.97% gc time, 4.43% compilation time)
[ Info: 2nd run
main(t): 14.817717 seconds (17.40 M allocations: 16.328 GiB, 2.70% gc time)
[ Info: 4-threads 1st run
MultiThread_main(t): 5.896782 seconds (18.30 M allocations: 16.973 GiB, 9.25% gc time, 3.88% compilation time)
[ Info: 4-threads 1st run
MultiThread_main(t): 5.175887 seconds (17.40 M allocations: 16.934 GiB, 1.61% gc time)
```

Benchmark seems to show no regression

tamasgal:
Sorry for coming late to the party... looks good good to me! 😃

PR #196, opened 2022-12-01
Fix show for LazyBranch

oschulz:


tamasgal:
Ah yes, I missed that ;)

PR #211, opened 2023-02-01
Add missing namespace before kOffsetP

Yuan-Ru-Lin:
Resolve https://github.com/JuliaHEP/UnROOT.jl/issues/210

### DIFF
--- a/src/root.jl
+++ b/src/root.jl
@@ -290,7 +290,7 @@ end
 function _normalize_ftype(fType)
     # Taken from uproot4; thanks Jim ;)
     if Const.kOffsetL < fType < Const.kOffsetP
-        fType - kOffsetP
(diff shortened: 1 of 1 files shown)


Moelf:
thanks!!

out of curiosity did you hit a bug with some files with this? I don't think we're using this, evident the fact that CI never caught it.

Yuan-Ru-Lin:
Yes, the bug was found when I was working on a proof-of-concept project in which I processed data files of the Belle II experiment. You can check it out at https://github.com/Yuan-Ru-Lin/BelleAnalysis.jl/blob/57f8b1a984177dfa76e83d76d1cf6b0c470dc78c/examples/reconstruct_B2Kpi.jl.

tamasgal:
Perfect,thanks! 🙂

PR #225, opened 2023-03-14
Match replace lazytree branch names

tamasgal:
The idea is to give the user the ability to rename branches. The current implementation of @Moelf is amazing, but sometimes the branch paths are long and annoying and result in awkward columns.

Before this patch:

```julia
julia> t = LazyTree(f, "E", [r"Evt/trks/trks.(dir|pos).([xyz])")
 Row │ Evt_trkstrksdir        Evt_trkstrksdir        Evt_trkstrksdir        Evt_trkstrkspos        Evt_trkstrkspos        Evt_trk ⋯     │ SubArray{Float6        SubArray{Float6        SubArray{Float6        SubArray{Float6        SubArray{Float6        SubArra ⋯─────┼─────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────
 1   │ [-0.487, -0.487, -0.4  [0.0369, 0.0369, 0.03  [-0.873, -0.873, -0.8  [125.0, 125.0, 70.7,   [615.0, 615.0, 585.0,  [446.0, ⋯
 2   │ [0.521, 0.521, 0.521,  [-0.175, -0.175, -0.1  [-0.835, -0.835, -0.8  [80.7, 80.7, 39.1, 33  [533.0, 533.0, 559.0,  [465.0, ⋯
 3   │ [-0.122, -0.122, -0.1  [-0.0817, -0.0817, -0  [-0.989, -0.989, -0.9  [194.0, 194.0, 96.5,   [593.0, 593.0, 581.0,  [457.0, ⋯
 4   │ [-0.23, -0.23, -0.23,  [-0.102, -0.102, -0.1  [-0.968, -0.968, -0.9  [204.0, 204.0, 124.0,  [590.0, 590.0, 571.0,  [440.0, ⋯
1 column omitted
```

After this patch:

```julia
julia> t = LazyTree(f, "E", [r"Evt/trks/trks.(dir|pos).([xyz])" => s"\1_\2"])
 Row │ pos_z                  dir_z                  pos_y                  dir_y                  dir_x                  pos_x   ⋯     │ SubArray{Float6        SubArray{Float6        SubArray{Float6        SubArray{Float6        SubArray{Float6        SubArra ⋯─────┼─────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────
 1   │ [125.0, 125.0, 70.7,   [-0.873, -0.873, -0.8  [615.0, 615.0, 585.0,  [-0.487, -0.487, -0.4  [0.0369, 0.0369, 0.03  [446.0, ⋯
 2   │ [80.7, 80.7, 39.1, 33  [-0.835, -0.835, -0.8  [533.0, 533.0, 559.0,  [0.521, 0.521, 0.521,  [-0.175, -0.175, -0.1  [465.0, ⋯
 3   │ [194.0, 194.0, 96.5,   [-0.989, -0.989, -0.9  [593.0, 593.0, 581.0,  [-0.122, -0.122, -0.1  [-0.0817, -0.0817, -0  [457.0, ⋯
 4   │ [204.0, 204.0, 124.0,  [-0.968, -0.968, -0.9  [590.0, 590.0, 571.0,  [-0.23, -0.23, -0.23,  [-0.102, -0.102, -0.1  [440.0, ⋯
 5   │ [58.6, 58.6, 30.1, 35  [-0.821, -0.821, -0.8  [546.0, 546.0, 565.0,  [0.54, 0.54, 0.54, 0.  [0.187, 0.187, 0.187,  [440.0, ⋯ 1 column omitted
```


Moelf:
I think we should try not to make `LazyTree()` any more complicated as it does many things already (including the custom type injection), we had:
- https://github.com/JuliaHEP/UnROOT.jl/issues/146

and I still think we should have a `rename()`, but as a separate API much like: https://dataframes.juliadata.org/stable/lib/functions/#DataFrames.rename!

tamasgal:
I thought it's minimal enough ;) the problem I see with the rename is that it requires a two step process and also carrying the information to two separate places while it could easily be done in one go.

If we do not need to do a lot of copying, we can also go for the rename option. I'll check that out. At least in my case, this rename thing is pretty normal and needed almost everywhere 🙈

tamasgal:
Another problem is that the user needs to know the normalised branch beforehand to create the rename map. This also means that changing the branch name normalisation is a breaking change for these awkward brach names, where you expect the user to do renaming.

Moelf:
>I thought it's minimal enough ;) 

eh, find another similar package that has the "main function" (defined as the one that most user will use to get the table-like thing back) handle renaming, I checked a few, and they don't usually do this:
- https://csv.juliadata.org/stable/reading.html#CSV.File
- https://dataframes.juliadata.org/latest/lib/types/#DataFrames.DataFrame

tamasgal:
Hm, I think I have to contradict, they do: `CSV` parsers usually offer "renaming" in the "main" function, like `header=` (https://csv.juliadata.org/stable/reading.html#header, https://pandas.pydata.org/docs/reference/api/pandas.read_csv.html) which can take a list. There is also the (sometimes optional) column name normalisation (https://csv.juliadata.org/stable/reading.html#normalizenames) which we also do, to get valid identifiers.

ROOT is of course much more different compared to CSV and handling complex, nested structures is way more complicated.

Just to regurgitate my example from above

```julia
t = LazyTree(f, "E", [r"Evt/trks/trks.(dir|pos).([xyz])")
```

will currently translate to the fieldnames

```
Evt_trkstrksdirx, Evt_trkstrksdiry, Evt_trkstrksdirz, Evt_trkstrksposx, Evt_trkstrksposy, Evt_trkstrksposz
```

while the original class which was serialised to this sub-branch is an `Evt`, which has fields named `vector<Hit>`, `vector<MCHit>`, `vector<Trk>`, `vector<MCTrk>`, `Header`, `vector<Weight>` and so on; the list goes on with dozens of other fields and the nightmare is that all the vectors of other classes are nested as well.

So I have to deal with the renaming of dozens of fields to make it more "usable" in the code and for the user.

If the renaming need to happen **after** the generation of these awkward names, one really needs to implement this by hand and rely on the normalisation procedure, which I consider to be something like an implementation detail (more or less). To mitigate this, the function to normalise branch names would need to be exported, in my opinion.

The code would need to be hardcoded if the renaming is external to `LazyTree`, something like (keep in mind that the list goes on with much more renaming, in my case):

```julia
t = LazyTree(f, "E", [r"Evt/trks/trks.(dir|pos).([xyz])")
new_t = rename(t,
    ["Evt_trkstrksdirx" => "dir_x",
    "Evt_trkstrksdiry" => "dir_y",
    "Evt_trkstrksdirz" => "dir_z",
    "Evt_trkstrksposx" => "pos_x",
    "Evt_trkstrksposy" => "pos_y",
    "Evt_trkstrksposz" => "pos_z",]
)
```

compared to this single line, where the information is minimal and it renames multiple branches in one go.

```julia
t = LazyTree(f, "E", [r"Evt/trks/trks.(dir|pos).([xyz])" => s"\1_\2"])
```

Of course, `rename()` could offer regex renaming as well, so that the user hooks up on the normalized names and create regex for those, but that does not feel good to me. The branch path is very well known, so that's the point where I naturally would hook in when I wanted to rename something, just like in CSV where the column names in the header line are known to the user (of course, in case of CSV one usually hooks into the column number due to the fixed ordering).

Apropos CSV. Taking your example with `CSV`-parsers, I think it would be fair to offer something like the "header"-renaming and then I don't see a big problem to give the user the option to also use regex 😉 or what do you think?

Maybe my cases are more like edge cases and other users of `UnROOT` are happy with the branchnames, but I have to mimic nested structures in order to allow event-by-event readout which materialise in Julia structs since our data is not really fitting the usual tabular style but is often doubly-jagged.

Anyways, let me know what you think.

tamasgal:
OK, I need some feedback here, I could not fix https://github.com/JuliaHEP/UnROOT.jl/issues/226 yet but I need to move on now and leave that for later.

My design process now heavily depends on whether we want this replacement feature or not 😉

src/iteration.jl
```diff
@@ -378,33 +378,45 @@ function LazyTree(f::ROOTFile, s::AbstractString, branches)
...
+    if length(v) >= 2 # only normalize name when branches are split
+        head = v[1]
+        tail = v[2:end]
+        # remove duplicated info
+        replace!(tail, head => "")
+        # remove known split branch information
+        replace!(tail, "fCoordinates" => "")
+        norm_name = join([head; join(tail)], "_")
+    end
+    @show s
```
Moelf, line 394: remove these debug prints?

Moelf:
I'm not against but please make sure CI passes and there's example and documentation

tamasgal:
OK thanks for the go! Of course I'll make everything pass and remove all the debug stuff (it's still draft) so that you can put your final stamp on it 😄

tamasgal:
OK, ready for review @Moelf 

I don't know what's wrong with `chaintrees()` on the nightly (see https://github.com/JuliaHEP/UnROOT.jl/issues/230), I could not reproduce it locally. Tried with the latest alpha and now compiling the laste `master` of Julia. I think it's not related.

tamasgal:
Tests pass on the latest `master`, so it's unrelated to the changes here.

docs/src/index.md
```diff
@@ -82,3 +82,59 @@ The laziness of the main interfaces are summarized below:
...
+julia> t.Evt_trkstrksposy
+10-element LazyBranch{SubArray{Float64, 1, Vector{Float64}, Tuple{UnitRange{Int64}}, true}, UnROOT.Nooffsetjagg, ArraysOfArrays.VectorOfVectors{Float64, Vector{Float64}, Vector{Int32}, Vector{Tuple{}}}}: 
+ [615.1089636184813, 615.1089636184813, 584.7490001284564, 582.9922451367319, 583.4532742304276, 583.212063675951, 583.182239372315, 582.5351568422853, 581.7743452806689, 583.6562661040083  …  597.4671030654561, 575.2131287145014, 582.7727094920472, 588.1060616488143, 574.8482883396676, 603.955888460846, 593.4457411811859, 576.6419859859786, 574.836340445788, 576.5382993955498]
+ ...
+ ...
+```
+
+Branch names are normalised so that they contain valid characters for
+identifiers. The branchname `Evt/trks/trks.pos.y` for example is therefore
+converted to `Evt_trkstrksposy`, which might be a bit inconvenient to use.
```
Moelf, line 123: this looks weird, I thought we would have `/ -> _` for each occurrence not just the first one, maybe it's a long standing bug

tamasgal: Ah OK, I thought that's intentional. Should we fix it here? It will be a breaking change though...

Moelf: we're going 0.9 -> 0.10 anyway xd

tamasgal: True ;)

Moelf:
looks good to me.

docs/src/index.md
tamasgal: Not sure about the fix. I just realise that it always replaces all occurrences of `head`, which might be not what we want (actually it could break the naming). We should probably only remove consecutive occurences of the same thing.

docs/src/index.md
```diff
@@ -82,3 +82,71 @@ The laziness of the main interfaces are summarized below:
...
+├─ E (TTree)
+│  └─ "Evt"
+└─ Header (Head)
+
+julia> t = LazyTree(f, "E", ["Evt/trks/trks.id", r"Evt/trks/trks.pos.[xyz]"])
+ Row │ Evt_trks_trks_i                     Evt_trks_trks_p                     Evt_trks_trks_p                     Evt_trks_trks_p      ⋯
+     │ SubArray{Int32,                     SubArray{Float6                     SubArray{Float6                     SubArray{Float6      ⋯
+─────┼───────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────
+ 1   │ [1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11  [615.0, 615.0, 585.0, 583.0, 583.0  [446.0, 446.0, 448.0, 448.0, 448.0  [125.0, 125.0, 70.7, ⋯
+ 2   │ [1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11  [533.0, 533.0, 559.0, 560.0, 559.0  [465.0, 465.0, 456.0, 452.0, 496.0  [80.7, 80.7, 39.1, 3 ⋯
```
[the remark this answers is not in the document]
tamasgal: yes, i was about to check the docs

docs/src/index.md
tamasgal: OK, fixed

docs/src/index.md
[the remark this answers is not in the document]
tamasgal: Done

Moelf (approved):

PR #231, opened 2023-03-30
Improves streamer logic for multiple jaggednes

tamasgal:
This adds support for triply jagged stuff, which is buried as `vector<something>` inside nested classes.

Currently only int32 and double is implemented and this needs a bigger refactoring, now that I understand a bit better  a few more edge-cases.

I'll go ahead with the merge since it does not affect the existing functionality. It's clear that the implementation is suboptimal but everything is typesafe, we just need to redesign the Jagg-type-system to e.g. use parametric types for the jaggedness or so.

With this patch, KM3NeT offline ROOT files (simulation and reco) are now fully readable ;)

```julia-repl
julia> f = UnROOT.samplefile("triply_jagged_via_custom_class.root");

julia> b = LazyBranch(f, "E/Evt/trks/trks.fitinf");

julia> typeof(b)
LazyBranch{SubArray{Vector{Float64}, 1, Vector{Vector{Float64}}, Tuple{UnitRange{Int64}}, true}, UnROOT.Offset6jaggjagg, ArraysOfArrays.VectorOfVectors{Vector{Float64}, Vector{Vector{Float64}}, Vector{Int32}, Vector{Tuple{}}}}

julia> b[2][3]
8-element Vector{Float64}:
   0.006795213995646554
   0.004663009290056574
 -78.11879123256927
  60.0
   0.0
   0.0
   2.164796367611156
  39.0
```


src/utils.jl
```diff
@@ -55,32 +55,33 @@ struct Nojagg      <:JaggType  end
 struct Nooffsetjagg<:JaggType  end
 struct Offsetjagg  <:JaggType  end
 struct Offsetjaggjagg  <:JaggType  end
+# this is a preliminary workaround for 6 byte offset jaggedness
```
Moelf, line 58: wtf?

tamasgal: Yeah sorry 🤣

PR #238, opened 2023-04-19
Recovering TTree baskets

tamasgal:
This PR implements the recovery mechanics of TBasket data.

### DIFF
--- a/src/bootstrap.jl
+++ b/src/bootstrap.jl
@@ -50,7 +50,7 @@ function unpack(io, tkey::TKey, refs::Dict{Int32, Any}, T::Type{RecoveredTBasket
     end
     fObjlen = size
     fNbytes = fObjlen + fKeylen
-    @warn "Found $(length(contents)) bytes of basket data (not yet supported) in a TTree."
+    @debug "Found $(length(contents)) bytes of basket data (not yet supported) in a TTree."
     RecoveredTBasket(contents, byteoffsets)
 end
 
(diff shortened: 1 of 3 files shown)


src/bootstrap.jl
```diff
@@ -2,7 +2,8 @@
 # dynamically in future.
 
 struct RecoveredTBasket
-    data
+    data::Vector{UInt8}
+    offsets::Vector{UInt32}
```
Moelf, line 6: this looks like the underlying data for a 1-jagged array?

Moelf:
interesting, so to put this mechanism in words, the recovery basket is the "last basket" in branch (makes sense, it's was being written when crash happened), which means:

1. we can't find it in `branch.fEntry`:
https://github.com/JuliaHEP/UnROOT.jl/blob/70732992b3ca6c47732ce9cea1ce1ef341672fe5/src/iteration.jl#L189-L191
2. but we will find it at the end of `branch.fBasket`:
https://github.com/JuliaHEP/UnROOT.jl/blob/70732992b3ca6c47732ce9cea1ce1ef341672fe5/src/iteration.jl#L56-L60

tamasgal:
So currently it's usually empty, but when there are recovered baskets, it's filled with `missing` (with the same number of of `missing` as there are "real" baskets) and at the end is a `RecoveredBasket`.

src/bootstrap.jl
tamasgal: that can be jagged or not, similar to the usual basket. If `offsets` is non-empty, it's jagged

Moelf:
ok! so concretely there should only be at most 1 real `fBaskets.elements` that's useful, this makes sense from ROOT writer perspective

PR #239, opened 2023-04-19
Index range access branch with recoveredtbasket

8me:
When accessing a branch using an index range there is still an error if a recovered basket is at the end. Trying to solve that problem ...

### DIFF
--- a/src/iteration.jl
+++ b/src/iteration.jl
@@ -466,10 +466,10 @@ function Base.getindex(ba::LazyBranch{T,J,B}, range::UnitRange) where {T,J,B}
     ib2 = findfirst(x -> x > (last(range) - 1), ba.fEntry) 
     if isnothing(ib1) #Check if we are completely on the recovered basket
         offset = ba.b.fBasketEntry[end]
-        iths = [-1]
+        iths = [-1] # use magic number -1 as address for recovered basket only
     elseif isnothing(ib2) # Check if we partially on the recovered basket
         offset = ba.fEntry[ib1-1]
-        iths = vcat(collect(ib1-1:length(ba.fEntry)-1), -1)
+        iths = vcat(collect(ib1-1:length(ba.fEntry)-1), -1) # append magic number -1 for recovered basket at the end of the basket address range
     else # Keep everything as it was
         offset = ba.fEntry[ib1-1]
         iths = ib1-1:ib2-1
--- a/src/iteration.jl
+++ b/src/iteration.jl
@@ -468,7 +468,7 @@ function Base.getindex(ba::LazyBranch{T,J,B}, range::UnitRange) where {T,J,B}
         offset = ba.b.fBasketEntry[end]
         iths = [-1]
     elseif isnothing(ib2) # Check if we partially on the recovered basket
-        offset = ba.fEntry[ib1]
+        offset = ba.fEntry[ib1-1]
         iths = vcat(collect(ib1-1:length(ba.fEntry)-1), -1)
     else # Keep everything as it was
         offset = ba.fEntry[ib1-1]
--- a/src/iteration.jl
+++ b/src/iteration.jl
@@ -462,11 +462,27 @@ function Base.iterate(tree::T, idx=1) where {T<:LazyTree}
 end
 
 function Base.getindex(ba::LazyBranch{T,J,B}, range::UnitRange) where {T,J,B}
-    ib1 = findfirst(x -> x > (first(range) - 1), ba.fEntry) - 1
-    ib2 = findfirst(x -> x > (last(range) - 1), ba.fEntry) - 1
-    offset = ba.fEntry[ib1]
+    ib1 = findfirst(x -> x > (first(range) - 1), ba.fEntry)
+    ib2 = findfirst(x -> x > (last(range) - 1), ba.fEntry) 
+    if isnothing(ib1)
+        ib1 = -1
+        offset = ba.b.fBasketEntry[end]
+    else
+        ib1 -= 1
+        offset = ba.fEntry[ib1]
+    end
+    if isnothing(ib2)
+        if ib1 < 0
+            iths = [-1]
+        else
+            iths = vcat(collect(ib1:length(ba.fEntry)-1), -1)
+        end
+    else
+        ib2 -= 1
+        iths = ib1:ib2
+    end
     range = (first(range)-offset):(last(range)-offset)
-    return ChainedVector(asyncmap(i->basketarray(ba, i), ib1:ib2))[range]
+    return ChainedVector(asyncmap(i->basketarray(ba, i), iths))[range]
 end
 
 _clusterranges(t::LazyTree) = _clusterranges([getproperty(t,p) for p in propertynames(t)])
--- a/src/iteration.jl
+++ b/src/iteration.jl
@@ -464,22 +464,15 @@ end
 function Base.getindex(ba::LazyBranch{T,J,B}, range::UnitRange) where {T,J,B}
     ib1 = findfirst(x -> x > (first(range) - 1), ba.fEntry)
     ib2 = findfirst(x -> x > (last(range) - 1), ba.fEntry) 
-    if isnothing(ib1)
-        ib1 = -1
+    if isnothing(ib1) #Check if we are completely on the recovered basket
         offset = ba.b.fBasketEntry[end]
-    else
-        ib1 -= 1
+        iths = [-1]
+    elseif isnothing(ib2) # Check if we partially on the recovered basket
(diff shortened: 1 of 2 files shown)


Moelf:
can we have benchmark? this is hot spot in our code

8me:
> can we have benchmark? this is hot spot in our code

It looks still a little bit messy, I just pushed to let you know I'm on that. I try to clean/optimize this later a little bit and then we can do some benchmarks :+1:

Moelf:
maybe hit it with 
```julia
for evt in tree
    evt.branch
end
```

`tree[range]` may have short cut

test/runtests.jl
```diff
@@ -263,7 +263,7 @@ end
         push!(bunches, sum(table.float_array[start:stop]))
     end
     testvalues = Vector{Float64}([528882.3f0, 1.5877059f6, 2.6465295f6, 3.705353f6, 4.764177f6, 5.823f6, 6.881823f6, 7.9406475f6, 8.999469f6, 1.0058294f7])
-    @test bunches == testvalues
+    @test bunches ≈ testvalues
```
Moelf, line 266: isn't the problem here you have `Float32` but your `bunches` are Float64?

tamasgal: Btw. I can reproduce the error on my machine. Johannes is currently investigating why he can't. Anyways, good catch with the Float32 stuff!

8me: Ok I'm kinda confused ... so what I do is:

1. Making sure I'm on the right commit, which is `commit a6719f3ecb8763a307571aec177630d326b339c4` according to `git log -1`
2. Checking my env is clean, which is according to `git status`
3. Switching inside docker env: `docker run -it --rm -v `pwd`:/test julia:1.7 /bin/bash`
4. There I run `julia --project=@. -e"using Pkg; Pkg.instantiate(); Pkg.add([\"MD5\",\"ThreadsX\"]); include(\"test/runtests.jl\")"` inside `/test`

and I get:
```
Test Summary:           | Pass  Total
LazyBranch and LazyTree |   26     26
```
:confused:

tamasgal: Whatever I do, I cannot reproduce, I always get 

```
LazyBranch and LazyTree: Test Failed at /Users/tamasgal/Dev/UnROOT.jl/test/runtests.jl:266
  Expression: bunches ≈ testvalues
   Evaluated: [528882.25, 1.587705875e6, 2.6465295e6, 3.705353e6, 4.764177e6, 5.8230005e6, 6.8818235e6, 7.9406475e6, 8.999469e6, 1.0058295e7] ≈ [528882.3125, 1.587705875e6, 2.6465295e6, 3.705353e6, 4.764177e6, 5.823e6, 6.881823e6, 7.9406475e6, 8.999469e6, 1.0058294e7]
```

But something with your "Docker-workflow" is not working either. The way you described it gives me other errors with Julia 1.7.

Do you have something weird in your startup files? As @Moelf pointed out, the `Float32` stuff is also a bit fishy, but that does not explain why it works for you. Maybe it's a Julia 1.7 thing? 

Anyways, I'd propose the good old `juila --project=.` (inside the top folder) and then `]test`. I use that all the time, everywhere 😉

8me: Regarding my workflow maybe it was a copy&paste thing that the \` arround pwd were not displayed correctly, so I changed to `$PWD`. Yeah running with `]test` reproduces the error, but I still find I weird that running `julia --project=@. -e"using Pkg; Pkg.instantiate(); Pkg.add([\"MD5\",\"ThreadsX\"]); include(\"test/runtests.jl\")"` which basically should do the same just lets the tests pass. Regarding something weird in the startup files: my home dir is not mounted to the docker and repo is clean. So there should not be something messy stuff going on in the background.

tamasgal: I am not sure but it might be that `MD5` and `ThreadsX` do something weird with the dependencies -- unexpected downgrades or whatever.

Moelf: `julia --project=@.` is wrong

tamasgal: Ah, I did not see the @ 😆

8me: Me neither, just reused from command history🤦🏻‍♂️

Moelf: well can you print `Pkg.status()` after `--project=.` and show us `git diff` or whatever?

8me: The `git diff` is empty and `julia --project=. -e"using Pkg; Pkg.status();"` shows:
```
Project UnROOT v0.10.6
      Status `/test/Project.toml`
  [1520ce14] AbstractTrees v0.4.4
  [65a8f2f4] ArraysOfArrays v0.6.1
  [c3b6d118] BitIntegers v0.2.7
  [5ba52731] CodecLz4 v0.4.0
  [ba30903b] CodecXz v0.7.0
  [6b39b394] CodecZstd v0.7.2
  [cd3eb016] HTTP v1.7.4
  [c8e1da08] IterTools v1.4.0
  [8ac3fa9e] LRUCache v1.4.1
  [9255714d] LibDeflate v0.4.1
  [3f54b04b] LorentzVectors v0.4.0
  [6fafb56a] Memoization v0.2.0
  [2a8e4939] Mixers v0.1.2
  [d96e819e] Parameters v0.12.3
  [08abe8d2] PrettyTables v2.2.3
  [91c51154] SentinelArrays v1.3.18
  [90137ffa] StaticArrays v1.5.21
  [09ab397b] StructArrays v0.6.15
  [bd369af6] Tables v1.10.1
  [9d84c17e] xrootdgo_jll v0.32.1+0
  [a63ad114] Mmap
```

Moelf: ??? why are you activating under `/test` folder?

8me: I added a `@test false` in order to make sure that this script is executed which worked like expected. So the behavior was like before: one method failed at the array comparison and the other method at the `@test false`. Anyways: I mixed something up, the type of the branch is `Float32` so the `sum` is `Float32` and written to the `Vector{Float64}` ... so I changed it at every spot to the correct type. Sorry for all the confusion :sweat_smile:

Moelf: >I added a @test false in order to make sure that this script is executed which worked like expected.

I'm not saying you're not executing the script `runtest.jl`, but you're using the UnROOT from your home environment.

The fact that `status()` doesn't show UnROOT.jl is worrisome -- what `UnROOT.jl` are you actually using?

src/iteration.jl
```diff
@@ -458,11 +458,20 @@ function Base.iterate(tree::T, idx=1) where {T<:LazyTree}
...
 
 function Base.getindex(ba::LazyBranch{T,J,B}, range::UnitRange) where {T,J,B}
-    ib1 = findfirst(x -> x > (first(range) - 1), ba.fEntry) - 1
-    ib2 = findfirst(x -> x > (last(range) - 1), ba.fEntry) - 1
-    offset = ba.fEntry[ib1]
+    ib1 = findfirst(x -> x > (first(range) - 1), ba.fEntry)
+    ib2 = findfirst(x -> x > (last(range) - 1), ba.fEntry) 
+    if isnothing(ib1) #Check if we are completely on the recovered basket
+        offset = ba.b.fBasketEntry[end]
+        iths = [-1]
```
Moelf, line 465: this destroys performance right?  allocation every single iteration?

Moelf:
can we have a even more basic test, that is not doing `branch[i:j]`?

why can't we just have `res = [evt.float_branch for evt in tree]`?

Moelf:
also why are we adding tests for `tree_with_large_array.root`?  does this file have recovered basket all along?

8me:
I think there is no sample file with a recovered tbasket which can be used that easy at the moment. So the idea was to add at least some tests for the `getindex` method for `UnitRange` for the standard use case. In make sure the added code doesn't break whats already there. @tamasgal suggested to sum up bunches inside the branch. As the sample files should be not replaced at some point, I thought some hardcoded values would be a nice test, because with that I don't rely on a other function, e.g. by creating the test values by accessing each index one by one.

Moelf:
I see. According to code coverage, we're not testing any of the added logic anyway, which is clearly the case until we have a sample file, we can wait I guess

tamasgal:
Yes, it's difficult to find a small file with recovered baskets. I was not able to produce one yet.

I think it's fine like this. @8me verified that the indexing works with one of our large files and I'd say it's good enough for now. The `-1` propagation is not the nicest approach but I want to work on the basket parsing mechanism anyways and will probably refactor that completely, now that I have more experience with all these structures.

We can go ahead with the merge, just need one or two lines of comments regarding `-1`.

PR #243, opened 2023-04-24
Fix type inference issue in basket access

tamasgal:
This PR separates the index calculation of normal baskets and recovered ones in two separate functions. The fix however for https://github.com/JuliaHEP/UnROOT.jl/issues/242 comes from the fact that now the return types are defined to be a `UnitRange{Int}` for both of them.

I did not manage to write a test to check the correct type inference yet.


tamasgal:
I opened an issue to remind (myself) of type stability tests: https://github.com/JuliaHEP/UnROOT.jl/issues/244

PR #245, opened 2023-04-29
Support TNtupleD

Yuan-Ru-Lin:
There was no support for `NTupleD` before i.e. `LazyTree("file.root", "ntupled")` will throw an `UndefVarError: TNtupleD is not defined`, as UnROOT parses the `fClassName` and found there is no corresponding `Symbol` defined in the module. I added the `Symbol` by defining a function `TNtupleD` that defers the call to `TNtuple`:

```julia
TNtupleD(io, tkey::TKey, refs) = TNtuple(io, tkey::TKey, refs)
```

which resolved the problem.

A corresponding testset, a script to generate the sample and the sample itself are included in the PR.


tamasgal:
Perfect, thanks!

PR #255, opened 2023-05-31
add thread locks to mitigate task migration

Moelf:
alternative to
-  #252

PR #256, opened 2023-06-01
improve correctness of thread locks

### LINKED ISSUE #251: Purge `threadid()`-based buffer from code base
Ever since Julia had task migration between OS threads, this practice has become technically a bug, we need to purge them.

Current idea is to use a `LRU` in place of the threaded() based buffer.

Moelf:
follow up of #255

fix #251

### DIFF
--- a/src/RNTuple/highlevel.jl
+++ b/src/RNTuple/highlevel.jl
@@ -89,14 +89,14 @@ Base.length(s::RNTupleSchema) = length(getfield(s, :namedtuple))
 function Base.getindex(rf::RNTupleField, idx::Int)
     tid = Threads.threadid()
     tlock = @inbounds rf.thread_locks[tid]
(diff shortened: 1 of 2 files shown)


Moelf:
I think we can, I can think of thing later

tamasgal:
No hurries 😉

Moelf:
basically the idea is you can see if the code works correctly under concurrency

```
julia> f(i) = (sleep(0.001); i);

julia> let state = zeros(Int, Threads.nthreads()), N=100
           @sync for i ∈ 1:N
               @async state[Threads.threadid()] += f(i)
           end
           sum(state), sum(1:N) 
       end
(100, 5050)
```

this is incorrect even if you only have 1 thread.

we can do this, and also do a version with `@spawn` instead of `@async`

PR #258, opened 2023-06-07
use task_local_storage for buffer

### LINKED ISSUE #257: How to test concurrency bug?
I believe we have fixed the issue in #256 but now I'm trying to test if previous version had bug, I have tried various versions of the following test:

```julia
julia> const t = LazyTree(UnROOT.samplefile("tree_with_large_array_lz4.root"), "t1");

julia> sum(t.int32_array)
4999950000

julia> function test(t)
           ch = Channel{Int}(length(t))
           @sync for i in eachindex(t)
               Threads.@spawn begin
                   put!(ch, t[i].int32_array)
               end
           end
           close(ch)
           sum(ch)
       end
test (generic function with 1 method)

julia> test(t)
```

using [v0.10.10] as a base, which should have a race condition bug. However, I keep getting the correct result:
```julia
julia> test(t)
4999950000

julia> test(t)
4999950000
```

I can force the bug if I make the following change:

```diff
diff --git a/src/iteration.jl b/src/iteration.jl
index 3ee5f7b..69739f4 100644
--- a/src/iteration.jl
+++ b/src/iteration.jl
@@ -180,6 +180,7 @@ function Base.getindex(ba::LazyBranch{T,J,B}, idx::Integer) where {T,J,B}
     localidx = if idx ∉ br
         _localindex_newbasket!(ba, idx, tid)
     else
+        sleep(0.0001)
         idx - br.start + 1
     end
     return @inbounds ba.buffer[tid][localidx]
```

which effectively induce a task migration. With this change, I can get garbage results, reliably:

```julia
julia> test(t)
3123750

julia> test(t)
1533490

julia> test(t)
1551910

julia> test(t)
2380810
```

after the fix PR, if I introduce the same change, I still get correct result:
```diff
diff --git a/src/iteration.jl b/src/iteration.jl
index f75adcc..bc68d35 100644
--- a/src/iteration.jl
+++ b/src/iteration.jl
@@ -179,6 +179,7 @@ function Base.getindex(ba::LazyBranch{T,J,B}, idx::Integer) where {T,J,B}
         localidx = if idx ∉ br
             _localindex_newbasket!(ba, idx, tid)
         else
+            sleep(0.0001)
             idx - br.start + 1
         end
         return @inbounds ba.buffer[tid][localidx]
```
(takes a lot of time...)
```julia
julia> test(t)
4999950000
```

Moelf:
~~performance to be tested.~~ it's fine

fix #257 , an alternative to #255 and #256


Moelf:
the error is due to https://github.com/apache/arrow-julia/issues/452

oschulz:
Hi @Moelf , not sure, but maybe [ContextVariablesX](https://github.com/tkf/ContextVariablesX.jl) could come in handy here as well?

Moelf:
the performance actually seems fine.......... nice?

Moelf:
let me think about this overnight and implement for RNTuple and implement the tests

Moelf:
I have verified the trick Stephen outlined works (would error out in older version), and doesn't hurt performance. I think this is good to go

Project.toml
```diff
@@ -27,7 +27,7 @@ Tables = "bd369af6-aec1-5ad0-b16a-f7cc5008161c"
 xrootdgo_jll = "9d84c17e-11f2-50ef-8cc9-e9701362097f"
 
 [compat]
-Arrow = "2 - 2.5, ^2.6.1"
+Arrow = "2 - 2.5"
```
tamasgal, line 30: sorry for my ignorance, but I guess 2.6 has a bug or so?

src/UnROOT.jl
```diff
@@ -19,6 +19,8 @@ using BitIntegers: @define_integers
 
 import Tables, PrettyTables
 
+_sleep_in_getindex() = false
```
tamasgal, line 22:
```suggestion
# Hook for task migration tests, see https://github.com/JuliaHEP/UnROOT.jl/issues/257
_sleep_in_getindex() = false
```

tamasgal (approved):
Looks good to me! The only suggestion is a tiny comment.

Project.toml
Moelf: yeah, https://github.com/apache/arrow-julia/pull/459

tamasgal:
I guess this will be v0.10.16 then?

tamasgal:
Sorry, accidentally requested Nick for reviewing 🙈

Moelf:
yeah idk, I'm dragging my feet on shipping this one because technically it leaks memory via the cache of primary Task.

@vchuravy 's best suggestion is for us to include a `UnROOT.empty_cache!()` function

tamasgal:
OK I guess I'll have a much closer look. I think if the cache is cleared each time a `ROOTFile` is properly closed, it's totally fine. If people leave dangling file pointers around, we cannot do much but offer a manual cache emptying mechanism like you mentioned.

Moelf:
No there's no dangling file pointer, just the last basket in each branch, on the main Task

Moelf:
maybe: https://blog.jling.dev/blog/thread_buffer_bug/

Moelf:
http://concurrencyfreaks.blogspot.com/2023/09/50-years-later-is-two-phase-locking.html?m=1

Turns out what we did in #255 has a name I believe

PR #261, opened 2023-06-11
improve streamer LRU

### LINKED ISSUE #260: LRU cache preventing unmapping of ROOT files
The LRU caches used for  the `(_)getindex` method `in root.jl` is preventing the unmapping of virtual memory from opened ROOT files, when the file is no more referenced by the user.

## Reproducing the issue

Following code can be used to reproduce the issue,

```julia
using UnROOT

# Set here path to the input file
const fname  = "Run2012BC_DoubleMuParked_Muons.root"

for i in 1:10
    println(i)
    f = ROOTFile(fname)
    t = LazyTree(f, "Events")
    sleep(2)
    f = t = nothing
    GC.gc()
    GC.gc()
    GC.gc()
end

GC.gc()

println("Press enter")
readline()
```

Look for the root file path in `/proc/<pid>/smaps` where pid is the julia process. Assuming it's the only julia process running on the machine, following command can be run:

```
cat /proc/$(pgrep julia | head -n 1)/smaps | grep root\$
```

Symptom: the file appears multiple times

Repeat the test after having removing the two `@memoize` statements from the `root.jl` file **or** after inserting in the test code a `Memoization.empty_cache!(UnROOT._getindex)` statement before the `GC.gc()` calls.

In both case the mapping will appear at maximum twice in  `smaps` in the course of the code execution showing the effectiveness of the unmapping during  garbage collection.

## Fix suggestion

Use directly LRUCache w/o the help of `@memoize` and make the caches member of the `ROOTFile` and `ROOTDirectory`  structs.  That way the memory addresses will be unmapped by the garbage collection of the ROOTFile and ROOTDirectory instances.

Moelf:
fix #260

### DIFF
--- a/src/UnROOT.jl
+++ b/src/UnROOT.jl
@@ -71,7 +71,6 @@ if VERSION >= v"1.9"
         show(devnull, t)
         show(devnull, t[1])
     end
-    Memoization.empty_cache!(_getindex)
 end
 
 end # module
--- a/src/bootstrap.jl
+++ b/src/bootstrap.jl
@@ -999,8 +999,9 @@ function TDirectory(io, tkey::TKey, refs)
     header_key = unpack(fobj, TKey)
     n_keys = readtype(fobj, Int32)
     keys = [unpack(fobj, TKey) for _ in 1:n_keys]
+    directory_cache = IdDict()
 
-    directory = ROOTDirectory(header_key.fName, dir_header, keys, fobj, refs)
+    directory = ROOTDirectory(header_key.fName, dir_header, keys, fobj, directory_cache, refs)
     return directory
 end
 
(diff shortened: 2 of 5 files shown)


Moelf:
@grasph I think this should fix it, but does it fix the overall memory issue you think you're seeing?

@tamasgal can you take a look to see if the merging of the two `_getindex()` is appropriate?

(I think maybe we should have more accessor function like `_getreference` and `_getdirectory`, instead of hard coding field names like what I'm doing right now)

tamasgal:
> @tamasgal can you take a look to see if the merging of the two _getindex() is appropriate?

looks good to me!

tamasgal (approved):

PR #266, opened 2023-06-29
fix normalize_ftype for array branches

gaede:
This fixes the handling of a branch that has been created by ROOT w/ type `std::array<float, 10> covMatrix{};` (from a ROOT file created w/ EDM4hep).
This branche's streamer has `ftype = 25` and after `_normalize_ftype()` ist has `ftype=-15` causing a crash w/ `Vector(empty)`.   After the fix the result is correctly `ftype=5 == Const.kFloat => Float32` and the branch can be read correctly.

I did not run any further tests, but checked that uproot does the same: https://github.com/search?q=repo%3Ascikit-hep%2Fuproot5%20kOffsetL&type=code.


tamasgal:
Ah perfect, thanks, this was obviously a typo. Can you maybe provide a small test sample?

Moelf:
oh wow, thanks a lot! Sample file would be good but I'm happy to merge this without sample since it's a simple typo.

Btw do you work on EDM4hep?

Moelf (approved):

gaede:
Hi,

I have created a small edm4hep file w/ 10 events and just the `PandoraPFO` branches: [https://syncandshare.desy.de/index.php/s/4RC22miYCZLTeMX](https://syncandshare.desy.de/index.php/s/4RC22miYCZLTeMX).
You can read this (after the fix) with
```
using UnROOT
f = ROOTFile("...path..../lcio2edm4hep/rv02-02.sv02-02.mILD_l5_o1_v02.E250_pandorapfo_10evt_edm4hep.root")
# -- create a lazy tree with all of these branches
t = LazyTree(f, "events", r"PandoraPFOs/PandoraPFOs.[(a-z)(A-Z))]"  )
evt = t[1]
#e = evt.PandoraPFOs_energy
cov = evt.var"PandoraPFOs_covMatrix[10]"
```
and see the covMatrix elements (790 for 79 particles.)  

There are still some auxiliary branches in the 'events' tree  that cannot be read by UnROOT (hence the regexp w PandoraPFO) as well as in the `podio_metadata` tree.

Any help/support in how these could be read would be highly appreciated.

 Cheer, Frank.

tamasgal:
Thanks for both of you `:)`

tamasgal:
> There are still some auxiliary branches in the 'events' tree that cannot be read by UnROOT (hence the regexp w PandoraPFO) as well as in the podio_metadata tree.

A separate issue would probably be a good idea, given the long list of UnROOT todos 🙈

PR #273, opened 2023-10-10
simple cache for ROOTFile and remove cache for TDirectory

### LINKED ISSUE #269: `LazyTree()` hang regression in 0.10.16
https://github.com/JuliaHEP/UnROOT.jl/compare/v0.10.15...v0.10.16

I see memory usage increasing when opening very wide tree (`nanoAOD` in this case)

I'm pretty sure it's related to the fact that we don't cache Streamer look up any more, we should find a way to add it back without re-introducing the LRU cache issue

### LINKED ISSUE #272: Performance for trees with a large number of branches
Performance for trees with large number of branches

Instantiating a `LazyTree` for a CMS nanoaod takes 20 seconds with v0.10.15,
which I expect too large for a good user experience. Below, the comparison with
`RDataFrame` used from ROOT REPL:

| Operation                  | Interpreted ROOT</br> TTree | ROOT</br> RDataFrame | UnROOT</br>  v0.10.15 | UnROOT</br> v0.10.16   |
|----------------------------|------------------|-------------------|----------|----------|
| Setup time [s]             |             0.10 |             0.003 |      21. | 115      |
| First-processing time [s]  |              2.9 |              0.08 |     0.21 |  0.25    |
| Second-processing time [s] |              2.9 |             0.004 |   0.0015 | 0.0015   |

The setup time corresponds to instantiating the `RDataFrame` or `LazyTree`, including file opening.

The first processing computes the sum of muons over all the events, the second
does the same for jets. The difference is due to the time of compilation happening at the first processing.

v0.10.16 is affected by an extra issue for the setup time already reported in PR#.

The Tree contains many branches (1689). The setup time increases non linearly
with the number of branches: it takes 1.9s if only 706 branches are included,
2.8s if only the other 983 branches are included, while as quoted in the table,
it takes 21s for the 706+983=1689 branches.

Note that the LazyTree memory footprint (from Base.summarysize()) is similar to
the file size: are the compressed baskets loaded in memory?

For v0.10.16, the footprint is 220 times the file size.

Testing code:
```julia
using UnROOT

@time l2 = LazyTree("nanoaod.root", "Events");
@time sum(l2.nMuon)
@time sum(l2.nJet)
println("Lazy Tree memory footprint: ", Base.summarysize(l2) , " bytes")
```

Moelf:
fix #272
fix #269

I looked at what we're actually indexing on a `ROOTFile`, and found out that in the case of
reading `NanoAOD`, it's just indexing `"Events"` for ~1400 times, and everything else once e.g.
(`Events/blah`). So I decided to just use a naive dict like this.

In the following PRs I will add more robust tests so we can figure out why we had a LRU cache
for `TDirectory` and if we need one in the future, and also prevent future regression.


Moelf:
@tamasgal you might want to fix the MD5 thing soon, apparently the internal API changed

tamasgal:
Oh, that MD5 comes out of nowhere. Thanks.

grasph:
Hello Jerry

On my side, tt stills take >=20 s (22s) for the setup time with master (3611ee)

LazyTree display is taking a very long time (I interrupted before end)

Moelf:
>On my side, tt stills take >=20 s (22s) for the setup time with master (3611ee)

do you have a sample file with not too many events?

grasph:
Thanks for looking at this. You can find a file with 10 events in https://cernbox.cern.ch/s/Vy6Yhz47vDaYiWa

grasph:
> Hello Jerry
> 
> On my side, tt stills take >=20 s (22s) for the setup time with master (3611ee)
> 
> LazyTree display is taking a very long time (I interrupted before end)

Clarification: with v0.10.16 setup time was 120s and it was v.10.15 it was 21s. So regression for the setup time is fixed. v0.10.15 and v0.10.16 didn't have the issue with the  display.

Philippe.

PR #277, opened 2023-10-10
add benchmark folder for local only

### LINKED ISSUE #127: Benchmarking for potential regression
We should make a benchmark suite to make sure we don't regress...

I don't think CI would be very happy to pull in a 2GB file for benchmarking ;) To start off with, it could even be just a local script to run.

Moelf:
fix #127 

I want to benchmark https://github.com/Moelf/UnROOT_RDataFrame_MiniBenchmark but obviously we can't do it on CI.

So what I'm probably going to do is to add the benchmark but if user doesn't have the file locally, print a warning

### DIFF
--- a/benchmark/README.md
+++ b/benchmark/README.md
@@ -1,0 +1,134 @@
+## How to use
+```bash
+> pwd
+~/.julia/dev/UnROOT/benchmark/ 
+
+> julia --project=.
+# ]instantiate
+```
+```julia
+julia> using PkgBenchmark
+
+julia> judge("UnROOT", "7f9eb14c270baefbc5b9da8d27530ac8b1a43975", "b59a632b835005ddc719a19211f99cf8f537114d")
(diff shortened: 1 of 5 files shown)


benchmark/README.md
```diff
@@ -0,0 +1,134 @@
...
+
+## Results
+A ratio greater than `1.0` denotes a possible regression (marked with :x:), while a ratio less
+than `1.0` denotes a possible improvement (marked with :white_check_mark:). Only significant results - results
+that indicate possible regressions or improvements - are shown below (thus, an empty table means that all
+benchmark results remained invariant between builds).
+
+| ID                    | time ratio                   | memory ratio                 |
+|-----------------------|------------------------------|------------------------------|
+| `["Latency", "load"]` | 0.02 (5%) :white_check_mark: | 0.03 (1%) :white_check_mark: |
```
Moelf, line 67: this is showing we have a 20x speed up

PR #286, opened 2023-10-13
Add a copy(::LazyBranch) method to prevent branch materialization on …

grasph:
For a LazyBranch br, Base.copy(br) was returning a vector instead of a LazyBranch and leading to uncessary read. This PR fix the issue.

### DIFF
--- a/src/iteration.jl
+++ b/src/iteration.jl
@@ -112,28 +112,38 @@ mutable struct LazyBranch{T,J,B} <: AbstractVector{T}
     buffer::Vector{B}
     thread_locks::Vector{ReentrantLock}
     buffer_range::Vector{UnitRange{Int64}}
+end
 
-    function LazyBranch(f::ROOTFile, b::Union{TBranch,TBranchElement})
-        T, J = auto_T_JaggT(f, b; customstructs=f.customstructs)
-        T = (T === Vector{Bool} ? BitVector : T)
-        _buffer = T[]
-        if J != Nojagg
-            # if branch is jagged, fix the buffer and eltype according to what
-            # VectorOfVectors would return in `getindex`
-            _buffer = VectorOfVectors(T(), Int32[1])
-            T = SubArray{eltype(T), 1, T, Tuple{UnitRange{Int64}}, true}
-        end
-        Nthreads = _maxthreadid()
-        return new{T,J,typeof(_buffer)}(f, b, length(b),
-                                        b.fBasketEntry,
-                                        [_buffer for _ in 1:Nthreads],
-                                        [ReentrantLock() for _ in 1:Nthreads],
-                                        [0:-1 for _ in 1:Nthreads])
+function LazyBranch(f::ROOTFile, b::Union{TBranch,TBranchElement})
+    T, J = auto_T_JaggT(f, b; customstructs=f.customstructs)
+    T = (T === Vector{Bool} ? BitVector : T)
+    _buffer = T[]
+    if J != Nojagg
+        # if branch is jagged, fix the buffer and eltype according to what
+        # VectorOfVectors would return in `getindex`
+        _buffer = VectorOfVectors(T(), Int32[1])
+        T = SubArray{eltype(T), 1, T, Tuple{UnitRange{Int64}}, true}
+    end
+    Nthreads = _maxthreadid()
+    return LazyBranch{T,J,typeof(_buffer)}(f, b, length(b),
+                                           b.fBasketEntry,
+                                           [_buffer for _ in 1:Nthreads],
+                                           [ReentrantLock() for _ in 1:Nthreads],
+                                           [0:-1 for _ in 1:Nthreads])
     end
 end
+
 LazyBranch(f::ROOTFile, s::AbstractString) = LazyBranch(f, f[s])
 basketarray(lb::LazyBranch, ithbasket) = basketarray(lb.f, lb.b, ithbasket)
 basketarray_iter(lb::LazyBranch) = basketarray_iter(lb.f, lb.b)
+
+"""
+    copy(x::LazyBranch)
+
+    Makes a shallow copy of x.
+"""
+Base.copy(x::LazyBranch{T,J,B}) where {T,J,B} = LazyBranch{T,J,B}(x.f, x.b, x.L, x.fEntry, x.buffer,
+                                                                  x.thread_locks, x.buffer_range)
 
 function Base.hash(lb::LazyBranch, h::UInt)
     h = hash(lb.f, h)
--- a/src/iteration.jl
+++ b/src/iteration.jl
@@ -141,7 +141,7 @@ basketarray_iter(lb::LazyBranch) = basketarray_iter(lb.f, lb.b)
 
     Makes a shallow copy of x.
 """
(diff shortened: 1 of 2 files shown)


Moelf:
I'm not so sure about this one. For example, `copy(reinterpret(...))` also changes type and return a Vector(). I think it could be argue that current behavior is the correct one already

Moelf:
## Arrow.jl

```julia
julia> d = Arrow.Table("/tmp/a.feather");

julia> d.x #this is memory-mapped
100-element Arrow.Primitive{Int64, Vector{Int64}}:
...

julia> copy(d.x)
100-element Vector{Int64}:
...
```

grasph:
According the Base.copy documentation it must "create a shallow copy of x: the outer structure is copied, but not all internal values."

Moelf:
but then you also have this from Julia itself:

```julia
julia> a = rand(2,2)'
2×2 adjoint(::Matrix{Float64}) with eltype Float64:
 0.651229  0.952613
 0.723043  0.00959459

julia> copy(a)
2×2 Matrix{Float64}:
 0.651229  0.952613
 0.723043  0.00959459
```

Moelf:
another thing is I just realized we no longer need `mutable struct LazyBranch`, it can be immutable.

But after making it immutable it breaks CI and seems like allocation in certain operation is increased somehow.

If we can make it immutable then clearly `copy` should just be a no-op. Except in this case we need clarification from others regarding `copy(<:AbstractVector)`.

grasph:
My idea was that the LazyBranch was conceptually immutable and this would avoid unnecessary materialization with DataFrame if copycols=false is not passed.

Looking at the DataFrames documentation, the behavior of Arrow.Table is to allow conversion of immutable to mutable columns:  "[to] get mutable columns from an immutable input table (like Arrow.Table), pass copycols=true explicitly." Changing the immutable nature of the object is arguable, although here it has some practical use here. The issue is that copycols=true is the default, which can be very expansive in HEP use case.

The case of  copy(A::Transpose) and  copy(A::Adjoint) is special. According to the documentation it "eagerly evaluate[s] the lazy matrix transpose/adjoint.". For sure return an object of the same type as the original is not wrong, it is changing the type which is an exception.

Moelf:
>The issue is that copycols=true is the default, which can be very expansive in HEP use case

But it is what it is, user of DataFrame needs to know what's default behavior. The problem of overriding copy is what user should do if they indeed want copycols=true?

I agree with your argument but I don't know if this is overly intrusive and confusing (i.e. what if user wants a mutable copy of a column?)

grasph:
Concerning the method `copy(<:AbstractVector)`, is is not documented. So I expect it can be considered as fallback implementation and should not be considered as providing the current behavior for all subtypes.

There is a Base.copymutable() function to turn an immutable into a mutable e.g, it turns a tuple to a vector.

Moelf:
I got this comment from @vtjnash so maybe we should just return the original object, what do you think?

image

src/iteration.jl
```diff
@@ -112,29 +112,38 @@ mutable struct LazyBranch{T,J,B} <: AbstractVector{T}
...
 basketarray_iter(lb::LazyBranch) = basketarray_iter(lb.f, lb.b)
 
+"""
+    copy(x::LazyBranch)
+
+    Makes a shallow copy of x.
+"""
+Base.copy(x::LazyBranch{T,J,B}) where {T,J,B} = LazyBranch{T,J,B}(x.f, x.b, x.L, x.fEntry, x.buffer,
+                                                                  x.thread_locks, x.buffer_range)
+
```
Moelf, line 146:
```suggestion
Base.copy(x::LazyBranch{T,J,B}) where {T,J,B} = x
```

grasph:
I think both options w and w/o copy of the buffer are valid. Independent buffer should provide better performance in case of concurrent access on two copies as it will prevent buffer invalidation, while no copy option minimizes memory allocations.

About this PR in general, since it is not clear what the best option is, I propose to:
  *  postpone the decision to specialize the copy function until we have a better view of the use cases;
  * document the possibility for the user to disable the materialization by defining himself the copy method as a no-op.

That way, we will keep the options opened. What do you think of this plan ?

Philippe.

Moelf:
that sounds good to me! 

I personally think `no-op` is okay, because I *think vast majority use cases are similar to when you want to do something in a different container type, e.g. `DataFrame`, the user probably stops interacting with the original `LazyBranch` variable.

Moelf:
>then that is a violation of the copy API, which states that operations on b do not affect a

ah ok we can't do `no-op`, so I think we know exactly what we can do in this case.

tamasgal:
This is really difficult since copying something lazy needs to decouple the structure from its laziness in order to fulfil what `copy` signals: having something which I can operate on without modifying the original thing.

We need to either break the laziness and materialise everything inside the structure (but then it's not a lazy thing anymore) or we create a completely different object with it's own references and independent caches, which sounds like a source of happy bugs 🙈

I don't know... I think changing the type is also a bit of a surprise, even though it makes sense in certain applications.

I agree that we should figure out what a user expects to be served when doing a copy of lazy branches.

src/iteration.jl
```diff
@@ -112,29 +112,38 @@ mutable struct LazyBranch{T,J,B} <: AbstractVector{T}
...
 basketarray(lb::LazyBranch, ithbasket) = basketarray(lb.f, lb.b, ithbasket)
 basketarray_iter(lb::LazyBranch) = basketarray_iter(lb.f, lb.b)
 
+"""
+    copy(x::LazyBranch)
+
+    Makes a shallow copy of x.
+"""
+Base.copy(x::LazyBranch{T,J,B}) where {T,J,B} = LazyBranch{T,J,B}(x.f, x.b, x.L, x.fEntry, x.buffer,
+                                                                  x.thread_locks, x.buffer_range)
```
Moelf, line 145:
```suggestion
Base.copy(x::LazyBranch{T,J,B}) where {T,J,B} = LazyBranch{T,J,B}(x.f, x.b, x.L, x.fEntry, deepcopy(x.buffer),
                                                                  x.thread_locks, copy(x.buffer_range))
```

vtjnash: A copy usually shouldn't call `deepcopy`, as that copies more than `copy` intendeds, and does so slowly

src/iteration.jl
```diff
@@ -112,29 +112,38 @@ mutable struct LazyBranch{T,J,B} <: AbstractVector{T}
...
 LazyBranch(f::ROOTFile, s::AbstractString) = LazyBranch(f, f[s])
 basketarray(lb::LazyBranch, ithbasket) = basketarray(lb.f, lb.b, ithbasket)
 basketarray_iter(lb::LazyBranch) = basketarray_iter(lb.f, lb.b)
 
+"""
+    copy(x::LazyBranch)
+
+    Makes a shallow copy of x.
+"""
+Base.copy(x::LazyBranch{T,J,B}) where {T,J,B} = LazyBranch{T,J,B}(x.f, x.b, x.L, x.fEntry, deepcopy(x.buffer),
```
Moelf, line 144:
```suggestion
Base.copy(x::LazyBranch{T,J,B}) where {T,J,B} = LazyBranch{T,J,B}(x.f, x.b, x.L, x.fEntry, copy.(x.buffer),
```

PR #290, opened 2023-10-24
add `Aqua.jl` tests and move `test/Project.toml`

Moelf:


test/Aqua.jl
```diff
@@ -0,0 +1,8 @@
+using Aqua
+
+@testset "Aqua.jl" begin
+    Aqua.test_all(UnROOT;
+    ambiguities = (; broken=true),
```
Moelf, line 5: all of the ambiguities come from upstream

test/Aqua.jl
```diff
@@ -0,0 +1,8 @@
...
+    piracy = (; broken=true)
```
Moelf, line 6: - https://github.com/JLTastet/LorentzVectors.jl/pull/11

trying to upstream here, but maybe we should just move everything into LorentzVectorHEP.jl

PR #296, opened 2023-11-29
Release 0.10.19

tamasgal:


tamasgal:
Let's go ahead with this, I think the autogenerated PR changelog is fine for now ;)

PR #299, opened 2023-12-20
Add support for more primitive (vector/jagged) leaf types

tamasgal:
Just a small PR which adds several vector/jagged primitive support for leaf types. This is a quick fix for https://github.com/JuliaHEP/UnROOT.jl/issues/298


tamasgal:
The reason for the incomplete coverage is the fact that I wanted to utilise some metaprogramming features to make it a bit more elegant and support all primitives... maybe this PR should be extended to cover them finally 🙈

tamasgal:
I added a few more...

PR #302, opened 2024-01-17
Remove nightly Julia build from CI

tamasgal:
The nightly build is often broken and the CI badge shows "failing" because it picks that up.
I don't think we need to waste resources on the nightly Julia CI anyways.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -23,11 +23,6 @@ jobs:
         os: [ubuntu-latest]
         arch: [x64]
         allow_failure: [false]
-        include:
-          - version: 'nightly'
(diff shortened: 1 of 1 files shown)


Moelf:
image


CI badge doesn't show nightly fail

tamasgal:
Currently not, but it shows it from time to time ;)

tamasgal:
Here is a short discussion: https://discourse.julialang.org/t/ignoring-nightly-failure-for-ci-badge/98028

But I am not sure if we really need nightly CI of Julia at all. I have not found the results of that job helpful at all. I always ignore that ;)

Moelf:
I kinda need it so that when new release come around it's not broken

PR #328, opened 2024-03-27
Restructure tests to allow running them individually

tamasgal:
Our test suite is already quite large (or let's say... slow 😉) and sometimes it's annoying to wait 1-2 minutes when working on something where it's sufficient to run just a part of the test suite during development.

This PR splits the test set into standalone scripts, so that they can be run individually.


tamasgal:
I always used the barebone `Test` package ;) Do you have a concrete suggestion?

Moelf:
maybe https://github.com/RelationalAI-oss/XUnit.jl

PR #333, opened 2024-04-02
[RNTuple] add test files for extension header

Moelf:
follow up of #332


src/RNTuple/fieldcolumn_reading.jl
```diff
@@ -110,7 +113,7 @@ function read_field(io, field::LeafField{Bool}, page_list)
 
     # pad to nearest 8*k bytes because each chunk needs to be UInt64
     bytes = read_pagedesc(io, pages, cr)
-    append!(bytes, zeros(eltype(bytes), 8 - rem(total_num_elements, 8)))
```
Moelf, line 113: `total_num_elements` is in unit of bits, so we need to do `rem` with 64 (bits), and then divide by 8 to convert to bytes

PR #337, opened 2024-05-06
Add TH3 support

tamasgal:
Not tested yet, but at least it does not error anymore when a TH3 is in a file ;)


tamasgal:
Oh dear.. This really killed my evening, I had to go deep into the rabbit hole.

Reading out some simple 3D histograms ended up with writing bootstrapping streamers for LaTeX strings because class inheritance is so cool.

I don't have a test file for the nasty files I have to deal with (no idea how to shrink them) but at least there is a test for a simple TH3F.

Anyways, there are now a bunch of new bootstrap streamers, maybe not 100% functional but at least we have something to improve in future.

We need a mechanism to combine fields from different structs btw. Initially I played around with some macros but hit the wall in the runtime scenario. Anyways, I think we already have lot of streamer code to think about a much nicer interface ;)

tamasgal:
I'll add 3D support to the `parseTH()` tomorrow. Merging now, since it does not break anything and only makes things slightly better 😆

Moelf:
nice!

Issue #336, opened 2024-05-04
[RNTuple] Roadmap of writing RNTuple to disk


Moelf:
My current plan is to break this huge task into four large stages:

## Stage 1 (Before July 2024)
In Stage 1 we aggressively use dummy data (raw bytes copied from real root files) and try to figure out what are the big-stroke steps we need when writing RNTuple to disk. Roughly they should be:
- TFile header, RNTuple Anchor, RNTuple header
- Update RNTuple anchor location
- Writing out pages
- Then RNTuple footer, TDictionary?
- Update locations in TFile header and TDictionary (RNTuple Anchor)

### Usability after Stage 1: 
bare minimal, user can write out primitive fields that are not too long.

## Stage 2 (Before Oct 2024)
In Stage 2 we try to peel off dummy stuff in two ways:
- Reduce amount of "hard code dummy and modify" code smell
- allow larger files by dealing with cluster group related things correctly

At this stage we will still be very rigid when it comes to schema complexity, but closer to production quality in the "file size" aspect.

### Usability after Stage 2: 
Basically functional for very simple data schema up to any size, as long as it fits in RAM (one-shot writing)

## Stage 3 (Before June 2025)
In Stage 3 we will expand the level of completeness in two critical ways:
- Improve schema support, in particular offset vector fields, consider possibility of switching to AwkwardArray.jl at this stage
- Allow appending / streaming of data onto disk

### Usability after Stage 3:
This will be analysis production ready -- users can write nanoAOD-kind of files with any size and can append to existing dataset, this will be the medium milestone for a production-level useful RNTuple writer

## Stage 4 (unknown time)
In Stage 4 we try to complete whatever is still missing, possible items:
- vastly improve schema supports, we need a good design for this to be maintainable in the long term
- introduce other features of RNTuple and provide APIs, such as alias column

Moelf:
![image](https://github.com/JuliaHEP/UnROOT.jl/assets/5306213/65edae5b-1b2e-4358-b0ca-3cc571f6ce7b)


I have fully mapped every byte in a file now. @tamasgal might be interested in this:

[hexproj.zip](https://github.com/user-attachments/files/15892767/hexproj.zip)

Moelf:
I expect them to show up after ROOT team freezes them this year/early next year. (1.0.0)

tamasgal:
We should add this to our documentation as the golden reference ;)

Moelf:
[...]

PR #343, opened 2024-06-30
[RNTuple] Writing Phase 1

Moelf:
This implements Phase 1 as outline in #336 

```julia-repl
julia> using UnROOT

julia> f = tempname()
"/tmp/jl_Vihin4X8H2"

julia> a = Dict("a" => rand(UInt32, 13));

julia> UnROOT.write_rntuple(open(f, "w"), a)

julia> LazyTree(f, "myntuple")
 Row │ a
     │ UInt32
─────┼────────────
 1   │ 4157448934
 2   │ 1517003214
 3   │ 3621726506
 4   │ 1460426740
 5   │ 3377398072
 6   │ 3689650254
 7   │ 1067622026
 8   │ 2595970883
 9   │ 681771139
 10  │ 2811723385
 11  │ 3599332316
 12  │ 418169821
 13  │ 4049611339


julia> LazyTree(f, "myntuple").a == a["a"]
true
```

- [x] Check C++ ROOT can read this

```julia-repl
julia> using UnROOT

julia> UnROOT.write_rntuple(open("a.root", "w"), Dict("a" => UInt32[0xcececece]))
```

```python
In [1]: import ROOT

In [3]: df =  ROOT.RDataFrame("myntuple", "a.root")
[ROOT.NTuple] Warning /build/jenkins/workspace/parallel_publishing_full_pipeline/build/projects/ROOT-HEAD/src/ROOT/HEAD/tree/ntuple/v7/src/RPageStorageFile.cxx:289 in ROOT::Experimental::Internal::RPageSourceFile::LoadStructureImpl()::<lambda()>:0: RuntimeWarning: Pre-release format version: RC 2

In [4]: list(df.Take['std::uint32_t']('a'))
Out[4]: [3469659854]
```

```julia-repl
julia> UnROOT.write_rntuple(open("a.root", "w"), Dict("abcd" => UInt32[0xcececece, 0xaebecede]))
```

```python
In [6]: df =  ROOT.RDataFrame("myntuple", "a.root")

In [7]: list(df.Take['std::uint32_t']('abcd'))
Out[7]: [3469659854, 2931740382]
```

```bash
$ root --version
ROOT Version: 6.33.01
Built for linuxx8664gcc on Jun 28 2024, 23:10:42
From heads/master@v6-31-01-2482-g50ec910e98
```


src/RNTuple/Writing/TFileWriter.jl
```diff
@@ -0,0 +1,565 @@
...
+function test_io(obj, expected; kw...)
+    a = IOBuffer()
+    rnt_write(a, obj; kw...)
+    ours = take!(a)
+    if ours != expected
+        color_diff(ours, expected)
+    end
+end
+
+function rnt_write(io::IO, x::UnROOT.FileHeader32)
```
tamasgal, line 76: Are these `legacy` flags something which should be propagated from the function call as keyword arguments?

tamasgal (approved):
Looks good to me! Looking forward to the future of writing ;)

src/RNTuple/Writing/TFileWriter.jl
Moelf: in this case we know `FileHeader32` will always be written out "legacy" (i.e. big endian) so we don't take keyword argument here.

Because it doesn't make sense to `rnt_write()` a `FileHeader32` with `legacy=false`.

Although maybe this should be called `bigendian=true|false`

peremato (approved):
It is difficult for me to review the technical changes. What surprises me is that the name  "myntuple" for the created RNTuple is implicit. 
I have tried to reproduce the C++ reading test but I failed because the `ROOT.RDataFrame` is not yet provided by ROOT.jl I think.

Moelf:
Yeah that's something I still have to figure out. Basically change that moves too many stuff at the moment so I haven't figured out what bytes to update etc.

Moelf:
still works as of:

```bash
$ root --version
ROOT Version: 6.33.01
Built for linuxx8664gcc on Sep 23 2024, 23:33:01
From heads/master@v6-31-01-3313-g2f5b71d8d9
```

PR #396, opened 2025-12-05
Move HTTP and XRootD related stuff into extensions

tamasgal:
The only part which I am stuck at right now is in `src/root.jl`:

```julia
function ROOTFile(filename::AbstractString; customstructs = Dict("TLorentzVector" => LorentzVector{Float64}))
    fobj = if startswith(filename, r"https?://")
        HTTPStream(filename)
    elseif startswith(filename, "root://")
        XRDStream(filename)
    else
        !isfile(filename) && throw(SystemError("opening file $filename", 2))
        MmapStream(filename)
    end
```

since both `HTTPStream` and `XRDStream` are defined in extensions, which causes a problem of course.

The question is then: how to deal with a type-matching based on string value with potentially undefined types.

For now I went with a symbol-based dynamic approach (which fails, but I will have a look asap):

```julia
    fobj = if startswith(filename, r"https?://")
        typesymbol = :HTTPStream
        !isdefined(UnROOT, typesymbol) && error("Opening HTTP streamed ROOT files requires to install and load the 'HTTP' module.")
        getproperty(UnROOT, typesymbol)(filename)
    elseif startswith(filename, "root://")
        typesymbol = :XRDStream
        !isdefined(UnROOT, typesymbol) && error("Opening XRootD streamed ROOT files requires to install and load the 'XRootD' module.")
        getproperty(UnROOT, typesymbol)(filename)
    else
        !isfile(filename) && throw(SystemError("opening file $filename", 2))
        MmapStream(filename)
    end
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -16,6 +16,41 @@
 UnROOT.jl is a reader for the [CERN ROOT](https://root.cern) file format
 written entirely in Julia, without any dependence on ROOT or Python.
 
+## Breaking API changes in v0.11.0
+<details><summary>XRootD and HTTP are now available via extensions</summary>
+<p>
+
+`UnROOT.jl` supports opening files remotely via XRootD which requires
+[`XRootD.jl`](https://github.com/JuliaHEP/XRootD.jl)
+and via HTTP using [`HTTP.jl`](https://github.com/JuliaWeb/HTTP.jl) seamlessly
+up to version 0.10.38.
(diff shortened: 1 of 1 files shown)


tamasgal:
Ah OK, I know what to do. I will convert that to actual function calls...

tamasgal:
OK, this will work. Ready @Moelf 

We could then do 0.11 because it's breaking.

tamasgal:
I will also add a README entry for the breaking change ;)

PR #401, opened 2026-03-21
Add RooFitResult object parsing

mmikhasenko:
This PR adds support for reading top-level `RooFitResult` objects in UnROOT.

The immediate objective is `RooFitResult` support, but in practice `RooFitResult` serialization depends on a slice of the broader RooFit `Roo*` object ecosystem. Because of that, this PR implements the minimum supporting `Roo*` readers needed to deserialize real RooFit results and expose the core fit payload from Julia.

With this change, UnROOT can read persisted `RooFitResult`s and recover:
- fit status scalars such as `status`, `covQual`, `edm`, `minNll`, and `numBadNLL`
- the constant parameter list when present
- the initial floating parameter list
- the final floating parameter list
- individual `RooRealVar` entries with names, values, and errors
- covariance and correlation matrices
- global correlation coefficients

### RooFit object support for `RooFitResult`
Added a new parser module in `src/roofit.jl` with support for:
- `RooFitResult`
- `RooArgList`
- `RooRealVar`

and the minimal helper machinery required to traverse the relevant RooFit object graph:
- selected RooFit base classes
- RooFit STL-backed collection structures used by `RooAbsCollection`
- `TMatrixTSym<double>` support for covariance/correlation data
- `TVectorT<double>` support for global correlation coefficients
- safe skipping of internal RooFit bookkeeping objects that are present in the stream but are not required for the validated result payload

### Integration into UnROOT object dispatch
Hooked the new parser into the package load path so top-level object access works through the normal:

```julia
f["object_name"]
```

interface.

### Regression coverage
Added dedicated RooFit regression coverage using:
- a real-world RooFit fixture in `test/samples/roofit_result_realworld.root`
- a synthetic RooFit fixture in `test/samples/roofit_result_synthetic.root`
- the C++ fixture generator in `test/samples/roofit_result_synthetic.C`

The tests now cover:
- top-level fit scalars
- parameter list lengths
- parameter lookup by name
- parameter values and errors
- covariance/correlation matrix extraction
- global correlation coefficient extraction
- result variants with and without covariance payloads
- display behavior for the new RooFit-facing structs
- coexistence with existing top-level `TVectorT<double>` support

### Documentation
Added:
- docstrings for the new RooFit-facing structs
- a user-facing docs page: `docs/src/roofit.md`
- docs navigation entry in `docs/make.jl`

The new documentation includes:
- how to open a `RooFitResult`
- how to inspect parameter lists
- how to access covariance/correlation data
- the current scope and limitations of the RooFit support

### Follow-up cleanup after review
After review, I also simplified a few parts of the implementation to better reuse existing UnROOT infrastructure:
- switched RooFit name/title parsing to the shared `nametitle(io)` helper
- reduced duplicated RooFit-local skip helpers
- factored `TVectorT<double>` payload reading into shared logic used by both RooFit parsing and the existing top-level object path

## Scope Clarification

This PR is intentionally scoped to making `RooFitResult` usable, not to implementing complete RooFit support in UnROOT.

`RooFitResult` cannot be decoded in isolation, so this change includes the subset of `Roo*` infrastructure needed to read the tested fixtures successfully. However, it should not be interpreted as full support for the general RooFit object ecosystem.

The implementation is focused on the validated, user-facing fit payload. Some RooFit bookkeeping structures are still treated as opaque internals and are skipped rather than exposed as Julia data structures.

## Validation

Validated locally with:

```bash
julia --project=. -e 'using Test, UnROOT; include("test/roofit.jl")'
julia --project=. -e 'using Test, UnROOT; include("test/roofit.jl"); include("test/misc.jl")'
```

This verifies both the RooFit-specific parsing paths and that the changes do not regress the existing top-level `TVectorT<double>` behavior.

## Possible follow-ups

Further work can be around broader hardening rather than core functionality:
- improve diversity of generated RooFit fixtures across more result shapes
- verify behavior against more ROOT/RooFit versions
- decide whether status-history decoding should be part of this PR or left for follow-up
- run the docs build in a fully provisioned environment

### AI usage
Implemented with `codex`, then reviewed and refined in follow-up cleanup passes.


tamasgal (approved):

mmikhasenko:
connected to https://github.com/JuliaHEP/UnROOT.jl/issues/162

PR #402, opened 2026-04-05
Remove Mixers and Parameters from deps

tamasgal:
Removes Mixers.jl and Parameters.jl from the deps


tamasgal:
Part of the clean-up plan in https://github.com/JuliaHEP/UnROOT.jl/issues/387

PR #403, opened 2026-04-05
Cleanup and unify bootstrapping and streamer intreface

tamasgal:
This is work in progress. I have a few local branches with different approaches to unify/simply the bootstrap and streamer interface and I'd like to collect them in a single PR. This was partly (re)triggered by https://github.com/JuliaHEP/UnROOT.jl/pull/401 where I realised that some historically grown interface designs are confusing, apparently even for AI tools 😉


tamasgal:
OK, I got plenty of issues fixed with this 😆 found also a bug in recovered basket reading and reworked also the TH handling. I love holidays with no travel and bad weather 😁

PR #438, opened 2026-06-12
[RNTuple] RNTuple fixing up to spec v1 and also complete writing

Moelf:
Human note:
- Most of the 2000 line changes are from adding a `.C` file to generate a big `.root` file, real changes are ~200 lines of code, which is good

Here's LLM's original summary:

```
You can now write a table out — validated by ROOT itself

  open("out.root", "w") do io
      UnROOT.write_rntuple(io, (x=[1.0, 2.0], s=["a", "b"]); rntuple_name="t")
  end

  The previous writer only produced valid files for the hardcoded default file name and a
  narrow set of types, because the TFile container records (key lengths, directory sizes,
  free-list, anchor) were canned byte blobs from one reference file. I rewrote the assembly
  in src/RNTuple/Writing/TFileWriter.jl to compute every record, and validated the output
  three ways: UnROOT round-trip, uproot 5.7.4, and ROOT 6.38's RNTupleReader (via cvmfs on
  this facility). All agree on the full type matrix — every integer width, Float16/32/64
  (Float16 needed a spec fix: it's a float field on a Real16 column, not std::float16_t),
  bit-packed Bool (previously wrote spec-invalid element counts padded to 64), unicode
  strings, vectors, and nested vectors, with arbitrary file/tuple names. Remaining
  limitations (documented in the docstring): single cluster, one uncompressed page per
  column, no struct/union columns.

  Reader bugs found by testing against the reference material

  Pulling uproot's test files (from scikit-hep-testdata, its test-file source) immediately
  exposed four reader bugs, all fixed and cross-checked against uproot:

  1. Zigzag decode used arithmetic shift (>>) instead of logical (>>>) — every split-int
  column with values of magnitude ≥ ¼ of the type range silently decoded to wrong numbers
  (e.g. 32767 → -1). Existing tests only used small values.
  2. SplitInt32 table entry claimed 64 bits (spec says 32).
  3. Deferred (extension) columns crashed with a BoundsError — clusters written before a
  column was added omit it from the page list. Reading now does per-cluster zero-padding
  keyed on first_ele_idx (this builds on the ColumnRecord field-order fix from earlier in
  the session).
  4. Real32Trunc/Real32Quant columns were unimplemented — added bit-unpacking,
  mantissa-truncation, and dequantization decoding, plus parsing of the value-range flag
  (0x02) in column records; all 11 columns of the float-types file now match uproot exactly.

  Test additions

  - 9 spec-v1.0 sample files (1.4–25 KB each) added to test/samples/RNTuple/, covering
  split-ints, deferred columns, trunc/quant floats, multicluster, nested structs, STL
  containers, bit columns, and the real-world staff ntuple — 39 new assertions in
  test/rntuple.jl, with expected values cross-checked against uproot.
  - New test/RNTupleWriting/highlevel.jl: round-trip tests for the full type matrix,
  non-default names, bit columns at 1/7/8/9/63/64/65/100 elements, and string edge cases.
  - Full suite: 1143 pass, 2 broken (pre-existing), 0 failures on 4 threads — this run also
  covers all of the earlier session's correctness fixes.

  Performance spot-check (from earlier audit, now measured)

  LazyTree iteration runs ~39 ns/row on a flat branch vs ~18 ns/row for an eager array()
  read — the gap is the per-element ReentrantLock that buys thread-safety; it's a deliberate
  design tradeoff and fine in practice. The only pathological case is many tiny baskets
  (16-event baskets → ~1.2 µs/row, dominated by per-basket type re-resolution), which real
  files don't hit.

  One known follow-up if you want it: suppressed columns (multiple column representations,
  uproot's test_1347 file) are still unsupported on read — the element-offset field after
  each inner page list is currently skipped.
```


src/RNTuple/Writing/TFileWriter.jl
```diff
@@ -526,115 +534,186 @@ function schema_to_field_column_records(table)
...
+    key = RBlob(; fNbytes = Int32(klen + length(payload)), fVersion = 4,
+                  fObjLen = Int32(length(payload)), fDatime = Stubs.WRITE_TIME,
+                  fKeyLen = klen, fCycle = 1,
+                  fSeekKey = Int32(position(file)), fSeekPdir = 100,
+                  fClassName = "RBlob", fName = "", fTitle = "")
+    rnt_write(file, key)
+    pos = position(file)
+    write(file, payload)
+    return pos
+end
```
Moelf: this is something we didn't do properly before.

RBlob is kinda not part of the RNTuple spec, basically this is a legacy (T-era, like TKey, TTree) ROOT file object to tell legacy ROOT machinary "here comes an R-era blob, like RNTuple".

src/RNTuple/constants.jl
```diff
@@ -47,8 +49,9 @@ RNTuple_ColumnType(type = 0x18, nbits = 32, name = :SplitReal32 , jltype = Float
 RNTuple_ColumnType(type = 0x19, nbits = 64, name = :SplitReal64 , jltype = Float64, issplit=true),
 RNTuple_ColumnType(type = 0x1A, nbits = 32, name = :SplitIndex32, jltype = Index32, issplit=true, isdelta=true),
 RNTuple_ColumnType(type = 0x1B, nbits = 64, name = :SplitIndex64, jltype = Index64, issplit=true, isdelta=true),
-# (0x1C, 10-31, :Real32Trunc  ), #??
-# (0x1D,  1-32, :Real32Quant  ), #??
+# variable-width: the actual bit width comes from the column record's nbits
+RNTuple_ColumnType(type = 0x1C, nbits = 32, name = :Real32Trunc , jltype = Float32, istrunc=true),
+RNTuple_ColumnType(type = 0x1D, nbits = 32, name = :Real32Quant , jltype = Float32, isquant=true),
```
Moelf, line 54: these are new types in RNTuple spec v1 that we didn't cover before.

Moelf:
Most of the `Stubs.xxx` variables are removed, and we're no longer faking to be ROOT author:

```
  ┌────────────────────────────────┬────────────────────────────────────────────────────┐
  │           Was a stub           │                        Now                         │
  ├────────────────────────────────┼────────────────────────────────────────────────────┤
  │ Stubs.WRITE_TIME (frozen       │ Real current time via new _root_datime() helper —  │
  │ 2024-11-13 10:02:19 on every   │ written files now carry the actual write timestamp │
  │ key)                           │  (verified: 2026-06-12 today)                      │
  ├────────────────────────────────┼────────────────────────────────────────────────────┤
  │ Stubs.magic_6bytes (0x40 0x00  │ Computed object preamble — (Const.kByteCountMask | │
  │ 0x00 0x42 0x00 0x02)           │  (2 + payload)) + class version, derived from the  │
  │                                │ anchor size                                        │
  ├────────────────────────────────┼────────────────────────────────────────────────────┤
  │ Stubs.dummy_padding1 (37 zero  │ Computed zeros(UInt8, 100 - position(file))        │
  │ bytes)                         │ zero-fill to fBEGIN                                │
  ├────────────────────────────────┼────────────────────────────────────────────────────┤
  │ Stubs.n_keys (0x00 0x00 0x00   │ Computed Int32(1) (the one anchor key)             │
  │ 0x01)                          │                                                    │
  ├────────────────────────────────┼────────────────────────────────────────────────────┤
  │ Stubs.file_preamble ("root" +  │ Explicit write(file, b"root") + Int32(63501)       │
  │ version blob)                  │ format version                                     │
  └────────────────────────────────┴────────────────────────────────────────────────────┘

  Deliberately left as Stubs constants (genuinely opaque/structural, both valid and
  name-independent, with comments explaining why):
  - tsreamerinfo_compressed — the zlib-compressed TList of TStreamerInfo describing
  ROOT::RNTuple; regenerating it would mean reimplementing ROOT's streamer serialization.
  - dummy_padding2 — the directory record's TUUID + reserved tail; the reader ignores it, so
  a zeroed UUID is sufficient.
```

src/iteration.jl
```diff
@@ -76,7 +76,20 @@ end
 function basketarray(f::ROOTFile, branch, ithbasket::AbstractVector{<:Integer})
     tuples = [rawbasketarray(f, branch, i) for i in ithbasket]
     rawdata = reduce(vcat, first.(tuples))
-    rawoffsets = reduce(vcat, last.(tuples))
+    # Each basket's offsets are relative to the start of its own data (first
+    # offset 0, last offset == content size), so when concatenating baskets we
+    # must shift every subsequent basket's offsets by the accumulated data size
+    # (same re-basing as `readbranchraw`).
+    if all(t -> isempty(t[2]), tuples)
```
Moelf, line 83: I wonder if this would fix some of the previously unsolved bugs related to dangling baskets

src/iteration.jl
```diff
@@ -160,13 +173,12 @@ basketarray(lb::LazyBranch, ithbasket) = basketarray(lb.f, lb.b, ithbasket)
 basketarray_iter(lb::LazyBranch) = basketarray_iter(lb.f, lb.b)
 
 function Base.hash(lb::LazyBranch, h::UInt)
+    # delegate to the TBranch/TBranchElement hash (file name, branch name,
+    # entry count): plain TBranch has no fClassName, and mutable read-state
+    # (buffer_range) must not enter the hash because `==` compares contents
```
Moelf, line 178: this seems like a bug fix from before.

src/root.jl
```diff
@@ -87,16 +90,17 @@ function ROOTFile(filename::AbstractString; customstructs = Dict("TLorentzVector
         unpack(head_buffer, FileHeader64)
     end
 
-    # Streamers
+    # Streamers (the record is exactly fNbytesInfo bytes; a fixed-size read
+    # would truncate large streamer records and over-read remote files)
     seek(fobj, header.fSeekInfo)
-    stream_buffer = OffsetBuffer(IOBuffer(read(fobj, 10^5)), Int(header.fSeekInfo))
+    stream_buffer = OffsetBuffer(IOBuffer(read(fobj, Int(header.fNbytesInfo))), Int(header.fSeekInfo))
```
Moelf, line 96: these are now more precise, before, we couldn't figure out how many bytes we needed to read

src/RNTuple/header.jl
```diff
@@ -45,19 +45,31 @@ struct ColumnRecord
...
+    ColumnRecord(type, nbits, field_id, flags, representation_idx, first_ele_idx, NaN, NaN)
 function _rntuple_read(io, ::Type{ColumnRecord})
     type = read(io, UInt16)
     nbits = read(io, UInt16)
     field_id = read(io, UInt32)
     flags = read(io, UInt16)
-    first_ele_idx = if !iszero(flags & 0x0008)
+    # final (1.0) spec: representation index comes before the optional first
+    # element index (deferred-column flag 0x01) and value range (flag 0x02)
+    representation_idx = read(io, UInt16)
```
Moelf, line 61: these are new features in v1 spec, basically you can "view" a column with a different "type", a bit like Julia's `reinterpret`.

src/streamsource.jl
```diff
@@ -61,5 +61,6 @@ function Base.seekstart(fobj::AbstractSourceStream)
 end
 
 function Base.read(fobj::AbstractSourceStream)
-    read(fobj, fobj.size - fobj.seekloc + 1)
```
Moelf, line 64: this looked like a bug before

tamasgal:
Holy shit 😆 ok i'll review

src/RNTuple/Writing/TFileWriter.jl
```diff
@@ -526,115 +534,219 @@ function schema_to_field_column_records(table)
...
-    rnt_header_obs = rnt_write_observe(file, rnt_header)
-    rntAnchor_update[:fNBytesHeader] = rnt_header_obs.len
-    rntAnchor_update[:fLenHeader] = rnt_header_obs.len
-    RBlob1_update[:fObjLen] = rnt_header_obs.len
-    RBlob1_update[:fNbytes] = rnt_header_obs.len + 34
+        zero(UInt64), rntuple_name, "", writer_identifier,
+        field_records, col_records,
+        UnROOT.AliasRecord[], UnROOT.ExtraTypeInfo[])
+    header_bytes = _buffer_bytes(io -> rnt_write(io, rnt_header))
+    fSeekHeader = _write_rblob(file, header_bytes, fdatime)
```
Moelf, line 678: this feels a bit more than just awkward... But for now we don't have some unified nested writing / tracking buffer size / close of mechanism.

tamasgal (approved):

PR #439, opened 2026-06-13
[RNTuple] add LZ4 compression for RNTuple writing and default to it

Moelf:
this goes on top of #438 (I don't have Stacked PR so no UI sugar :()


tamasgal:
Let's go ;)

tamasgal (approved):

PR #458, opened 2026-08-26
Support ArraysOfArrays v1

oschulz:
ArraysOfArrays v1.0.0 renames `VectorOfVectors` to `PartsView` (the old name remains as a deprecated alias) and adds the element type as a fifth type parameter, so the four-parameter `VectorOfVectors{T,VT,VI,VD}` spelled out in `_field_output_type` is no longer a concrete type on v1, which breaks the type assertions in `read_field` for struct fields containing vectors.

* `PartsView` is used throughout, aliased to `VectorOfVectors` on ArraysOfArrays v0.6, which stays supported.
* `_field_output_type` for vector fields goes through a small version-conditional helper that spells out the concrete type, on v1 including the element type (computed the same way ArraysOfArrays does).
* Compat `ArraysOfArrays = "^0.6, ^1"`, docs snippet and patch version bump.

Verified locally on Julia 1.12 with ArraysOfArrays v1.0.0 by reading all columns of the RNTuple STL-container and nested-struct samples (output types are concrete and identical to `_field_output_type`) and the jagged TTree samples.

Created by generative AI.


oschulz:
Superseded by #459: `VectorOfVectors` stays a supported alias in ArraysOfArrays v1.0.1, so the change shrinks to the output-type helper.

PR #459, opened 2026-08-26
Support ArraysOfArrays v1

oschulz:
Adds compatibility with ArraysOfArrays v1 while keeping support for v0.6.

ArraysOfArrays v1 adds the element type of `VectorOfVectors` as a fifth type parameter, so the four-parameter `VectorOfVectors{T,VT,VI,VD}` spelled out in `_field_output_type` is no longer a concrete type there, and the type assertions in `read_field` for struct fields containing vectors would fail. The vector field output type now goes through a small version-conditional helper that includes the element type on v1 (computed the same way ArraysOfArrays does). Everything else UnROOT uses, including `VectorOfVectors` itself, is unchanged in v1.

Verified on Julia 1.12 with ArraysOfArrays v1.0.0 and v0.6.6 by reading all columns of the RNTuple STL-container and nested-struct samples (output types concrete and identical to `_field_output_type`) and the jagged TTree samples.

Compat `ArraysOfArrays = "^0.6, ^1"`, patch version bump.

Created by generative AI.


oschulz:
@tamasgal , @Moelf : Verified manually, ~~should be good to go.~~

oschulz:
Actually, hold off on this, I think I can do this more elegantly.

oschulz:
@tamasgal , @Moelf : We actually have to do it as it is, at least for now, I'd like to keep ArraysOfArrays v0.6 compatibility in UnROOT for a while.

The `Base.promote_op(view, VT, UnitRange{Int})}` looks a bit ugly, but this is essentially low-level code, and it's the price for having correct element type in `VectorOfVectors`. I'll see what functionality I can add to ArraysOfArrays in the future, if more code using ArraysOfArray needs to do this, but I'd say let's merge this PR as is now.

oschulz:
@tamasgal will you merge/release or should I?

Moelf:
I mean the tests are red...

oschulz:
> I mean the tests are red...

Ooops, sorry, on it.

oschulz:
CI passes now, except for Downgrade. Claude says:

The Downgrade job's failure is unrelated to this PR. It fails in Aqua's `test_stale_deps`
because julia-downgrade-compat pins the test-only dependency Aqua to its oldest compatible
version, 0.8.0, whose stale-deps check misreports the test-target extras (Arrow, DataFrames,
HTTP, XRootD, Aqua itself) as unloaded package deps. The same job fails identically on every
recent push to `main`, e.g.

- https://github.com/JuliaHEP/UnROOT.jl/actions/runs/32200580427 (2026-08-19)
- https://github.com/JuliaHEP/UnROOT.jl/actions/runs/31560905865 (2026-08-12)
- https://github.com/JuliaHEP/UnROOT.jl/actions/runs/30935710010 (2026-08-04)

whereas the regular test jobs with Aqua 0.8.16 pass the check. Either of these one-line
changes would fix it, in a separate PR:

- add `Aqua` to the `skip:` list of `julia-downgrade-compat` in `.github/workflows/downgrade.yml`
  (test-only deps don't need downgrade testing), or
- raise the lower bound in `[compat]` to `Aqua = "0.8.16"` (or whichever earlier patch fixed it).

tamasgal:
OK let's go, thanks ;)

oschulz:
Thanks!

After ArraysOfArrays v1.0 has been around for a while, we can drop v0.6 support in UnROOT and see which of the new AoA features we can use.

